PR #2, opened 2016-04-29
count number of retries

alexpantyukhin:


rogeralsing:
I think we need to keep the restart stats outside of the PID, in a map in the parent actor or something.
PID's are passed across nodes so the information doesnt really belong in the PID itself.

PR #19, opened 2016-11-04
interaction_test: fixed the actor type

samanbarghi:
Become and Unbecome echo tests were using the wrong actor.


rogeralsing:
Thanks!

PR #21, opened 2016-11-13
Prevent actor cells creation under same ID

dtravin:
Hi, @rogeralsing!

The project lacks the API to 'lookup or create" actor and spawning actors under same ID does not give any error.
Have a look at the test and fix


rogeralsing:
looks good! there is some merge conflict though. can you have a look at that?

rogeralsing:
Thanks!

PR #23, opened 2016-11-15
Actor passivation on inactivity timeout implemented as plugin

dtravin:
Hi @rogeralsing 
Lifecycle of an actor is not finite unless stopped explicitly.
Stopping actor after configurable timeout is implemented as a plugin
Pls check

### DIFF
--- a/actor/actorref.go
+++ b/actor/actorref.go
@@ -4,6 +4,7 @@ import "log"
 
 type ActorRef interface {
 	Tell(message interface{})
+	Ask(message interface{}, sender *PID)
 	SendSystemMessage(message SystemMessage)
 	Stop()
 }
@@ -19,7 +20,11 @@ func NewLocalActorRef(mailbox Mailbox) *LocalActorRef {
 }
 
 func (ref *LocalActorRef) Tell(message interface{}) {
-	ref.mailbox.PostUserMessage(message)
+	ref.mailbox.PostUserMessage(UserMessage{message: message})
+}
+
+func (ref *LocalActorRef) Ask(message interface{}, sender *PID) {
+	ref.mailbox.PostUserMessage(UserMessage{message: message, sender: sender})
 }
 
 func (ref *LocalActorRef) SendSystemMessage(message SystemMessage) {
@@ -45,6 +50,10 @@ var deadLetter ActorRef = new(DeadLetterActorRef)
 
 func (DeadLetterActorRef) Tell(message interface{}) {
 	log.Printf("Deadletter got %+v", message)
+}
+
+func (DeadLetterActorRef) Ask(message interface{}, sender *PID) {
+	log.Printf("Deadletter was asked %+v", message)
 }
 
 func (DeadLetterActorRef) SendSystemMessage(message SystemMessage) {
--- a/actor/bounded_mailbox.go
+++ b/actor/bounded_mailbox.go
@@ -7,17 +7,22 @@ import (
 	"github.com/Workiva/go-datastructures/queue"
 )
 
+type UserMessage struct {
+	message interface{}
+	sender *PID
+}
+
 type boundedMailbox struct {
 	throughput      int
 	userMailbox     *queue.RingBuffer
 	systemMailbox   *queue.RingBuffer
 	schedulerStatus int32
 	hasMoreMessages int32
-	userInvoke      func(interface{})
+	userInvoke      func(UserMessage)
 	systemInvoke    func(SystemMessage)
 }
 
-func (mailbox *boundedMailbox) PostUserMessage(message interface{}) {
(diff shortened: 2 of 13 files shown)


rogeralsing:
Look great, but two things that I would like to discuss here.

1) `Ask` needs to work with remoting support too if that is to be introduced. so UserMessage needs to be a proto message in that case.

Another issue with the `Sender()` approach is as it mutates between message receives, if you use the `context` inside a closure that runs in e.g. a go routine. for example if the actor spins up some async process on the side. then Sender() may or may not be avaiable or even point to the wrong actor if the async process continues when the actor receives a new message.

We have seen this countless of times in Akka.NET, which was one of the reasons why I opted for explicit sender properties in messages in GAM.

2) Do you think the passivisation support could be done without having the embedded struct?
e.g. bind the state in a closure and somehow utilize that in the plugin?

The reason why I would preffer this is that in the Cluster Virtual Actor support, we could simple write normal actors and have them passivated by default.
e.g. by letting the cluster support add this plugin on the actor props.
Do you think this would be doable?

rogeralsing:
The Remoting `MessageEnvelope` can probably be extended to carry the sender pid, that way, any changes would have minimal impact on the remoting layer,
We would just have to be able to pick the sender from the user message and put it in the MessageEnvelope and unwrap it on the other side.

That would work

rogeralsing:
Another thing, the UserMessage will double the amount of allocations and put pressure on the GC.
What do you think if we instead of having a special message, just having an interface with a sender member?
So if the message has a sender of type *PID then it is compatible with the interface. and we could use that to extract the same information and store in the context.
w/o any extra allocations, and zero changes to remoting layer.

dtravin:
1) a) I did not rewrite remoting part yet, because iam not 100% fluent there. I think we both understand that including infrastructural information into business event is a bad design. It might be in an envelope header or metadata, but not the body that is persistable. So, I made an envelope. Yes this is one more allocation, but there is no other way to attach metadata to message.
b) Regarding `Sender()`, Akka has the same API `self.sender()` on [http://doc.akka.io/api/akka/2.0/akka/actor/ActorContext.html](Actor) transparent to the message. 
We can do nothing about the case when application holds reference to ActorRef and sends message using that later. Instead of that application should look up and actor by identifier in place when communication happens.

2) yes. In akka they have another approach that allows application to set receive timeout and handle it in Receive like this (taken from akka docs) `case ReceiveTimeout ⇒ context.parent ! Passivate(stopMessage = Stop)`
It looks more sophisticated. Reference to actor is getting to some sort of scheduler structure and it has to be maintained properly. Complicated, isnt it ?

rogeralsing:
I will merge this, I have run some benchmarks with extra allocations per message and there was no visible performance change from what I can see.

Also the Sender() method is really not different from the Message() method in terms of concurrency/mutability problems, so I think its fine.

dtravin:
Double-checked the ReceiveTimeout from akka source.
In short, they use system.scheduler() and pass self to it and ActorCell has a field that holds a reference to cancel or reset the timer.
So anyway an embedded structure is required.

rogeralsing:
Could you take a look at the `endpointWriterMailbox` in the remoting layer?
It's not quite compatible with the new changes.

The endpointWriterMailbox works by taking all messages it can in one big go from the backing queue, to do a sort of batch processing of messages.
This is done so that when the endpoint writer is activated, it takes all messages it can, and then wrap them in one big envelope to push over the wire, to minimize the number of individual network writes.

the UserMessage members are however private, so I'm not sure how to build an external mailbox with the latest changes

rogeralsing:
I've managed to fix the broken bits in the remoting layer.
Everything is up and running again.
There is a bit of performance regression in remoting now, but most of it seems to come from the additional Sender field on the Message Envelope.

There are a few design issues that we should try to fix here.
There is a bit of odd mis match between UserMessage and MessageEnvelope they are sort of the same, yet different.. this might be possible to make nicer.
e.g. maybe UserMessage should be serializable and sent as the payload w/o unwrapping before sending.

Also, regarding the `PID.Ask` maybe that should be made nicer too?
It could create the RequestResponsePID internally so that the API could be nicer.
e.g.

res := myActor.Ask(message{}).ResultOrTimeout(..)

that would be better I think.

dtravin:
1) MessageEnvelope is a protocol level structure independent from what is happening in actor context.
I dont see anything wrong with this, but I was thinking to unite UserMessage and SystemMessage because it is totally ok for a system message to have a sender

2) It is a good idea to have PID.Ask to return *PID

I still have no clear picture about remoting here. Too many corner cases uncovered.
Like map eviction not implemented [https://github.com/AsynkronIT/gam/blob/dev/remoting/endpoint_manager.go#L21](here)

PR #24, opened 2016-11-16
Better API for PID.Ask

dtravin:


rogeralsing:
Is there a reason why the result values are reutrned as `error, response` instead of `response, error`?
I have never seen that order before, but then again, I started Go this year so it might just be that.

rogeralsing:
Also, would it make sense to use the `context` package to provide the timeout support for `Ask` ?
https://golang.org/src/context/context.go?s=13917:13994#L426

dtravin:
"By convention, errors are the last return value and have type error, a built-in interface."
My bad

dtravin:
Give me an example API you want to achieve with context.
Are you trying to hide the deferred Stop() from application code in such way ?

rogeralsing:
I was mostly thinking that there might be some more idiomatic way to handle timeouts in Go.
Just so that GAM don't introduce some made up concept (Response) if there is a better way to handle it.

btw. I have adapted the tests to use the new Ask API, love it! much more straight forward now.

PR #26, opened 2016-11-18
ProcessRegistry: unlock directly without defer

samanbarghi:
Using defer has known performance issues when used along with
unlock: https://github.com/golang/go/issues/14939

Removing defer from process_registry unlock makes the library run 50%
faster.


rogeralsing:
Thanks, I did not know that this :)

Issue #27, opened 2016-11-20
Make an option to subscribe for PID.Stop()


dtravin:
In tests there is need to wait until actor is fully stopped.

rogeralsing:
The PR have been merged.

PR #28, opened 2016-11-20
#27 PID.StopFuture() implemented

dtravin:
Implemented StopFuture using context.Watch


rogeralsing:
Love it, I have tests where this is needed

PR #29, opened 2016-11-22
Replaced hashmap + rwlock with ctrie in ProcessRegistry

samanbarghi:
With ctrie the performance is 2x better in comparison with map+lock; for scenarios where the contention is high.

### DIFF
--- a/actor/process_registry.go
+++ b/actor/process_registry.go
@@ -1,12 +1,10 @@
 package actor
 
 import (
-	"log"
 	"strconv"
-	"sync"
 	"sync/atomic"
 
-    "github.com/Workiva/go-datastructures/trie/ctrie"
+	"github.com/Workiva/go-datastructures/trie/ctrie"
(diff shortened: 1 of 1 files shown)


rogeralsing:
Awesome, thanks!

rogeralsing:
I am seeing some serious performance regression when looking up PID's with this change.
The Remote Benchmark dropped from 1.1 mil to 9 00k messages per sec, which leads me to beleive the regression is even worse if there is no network in between.

I get that this implementation might be faster in the end with high contention, but we need to have solid numbers for that first.
I will revert this for now, and re-introduce it once I can verify that this yields a better overall performance profile.

samanbarghi:
I see, yes probably a wider performance tests should be made for each change to figure whether there is any side-effects on other workloads. For high contention scenarios ctrie scales very good. Btw, what are you using to measure  remote performance?

PR #30, opened 2016-11-22
UnboundedMailbox: simple fix to avoid goroutine overhead

samanbarghi:
To avoid destroying and recreating goroutines for handling messages,
in cases where the inter-arrival time of messages is not short
enough or long enough to prevent the goroutine from exiting, it is
better to simply wait and check the queues a few time before letting
go of the goroutine. This adds a bit of CPU utilization and possibly
small memory overhead, but boost the performance for those corner
cases almost 3 times.

It can be changed to a configurable variable later to be tuned depending 
on the workload.


rogeralsing:
I like it!, thanks

rogeralsing:
This had some *really* nice effects on the remote benchmark.
Up from ~1 mil to 1.2 mil messages per sec.

Running the same with `GOGC=off` yields 1.6+ mil messages per sec which is pretty rad :)

samanbarghi:
That's great! :+1:

PR #32, opened 2016-11-23
Consisten router and remote group

pavelsmejkal:
Hi, i was playing with gam little bit to allow routing of remote messages to using any routing strategy. Also included consistent router. Can you have a look if it is something that gam project would benefit from?

see examples/remoterouting/client/main.go how api looks like


remoting/remote_group.go
```diff
@@ -0,0 +1,94 @@
...
+	self.destinationProducer(context, router)
+}
+
+func CreateDestinations(context actor.Context, service string, hosts []string) []*actor.PID {
+	pids := make([]*actor.PID, len(hosts))
+
+	for i, h := range hosts {
+		pid := actor.NewPID(h, service)
+		// Unfortunately this doesn't work since router hasn't been registred yet
+		//context.Watch(pid)
```
rogeralsing: Is the problem here that the actor is booting up and have not yet been registered in the process registry?

rogeralsing:
I love it!

actor/hashing/hashring/hashring.go
```diff
@@ -0,0 +1,30 @@
+package hashring
+
+import (
+	"github.com/AsynkronIT/gam/actor"
+	"github.com/serialx/hashring"
```
rogeralsing: I didn't know about this lib,
It should be possible to use this to replace the home baked hashring I do here, right?
https://github.com/AsynkronIT/gam/blob/dev/cluster/registry.go#L28

rogeralsing:
Merged! thanks

actor/hashing/hashring/hashring.go
pavelsmejkal: I can create default hasher that will using this home bakend hashring.

remoting/remote_group.go
pavelsmejkal: I was bit wrong. Problem is that remote actors can't receive system message. Needs to dig more deep to get idea of watching.

rogeralsing: Ah, it's not fully implemented yet.
the remote_actorref.go needs to support sending system messages, and the remoting layer also needs to be able to handle this.

The system messages also needs to be ported to protobuf in order to do this.

rogeralsing: I'm working on this now. watch, unwatch and terminated are now protomessages

PR #40, opened 2016-12-14
delay spawn until services are started

stuartcarnie:
ensures these aren't spawned in unit tests or create other unintended side effects when importing these packages


rogeralsing:
Looking good!

PR #42, opened 2016-12-16
simplify processMessages loop

stuartcarnie:
* fixes issue that a suspended mailbox will loop forever if it is not resumed
* early exit if a mailbox is suspended and no more system messages are consumed
* on exit, if new messages are available, continue processing more
    * avoids call to schedule which will start a new goroutine immediately


rogeralsing:
Looks good, merging

stuartcarnie:
👍

PR #43, opened 2016-12-19
fix race condition

stuartcarnie:
* revert to original logic, with minor optimization to restart
  message processing if atomic CAS succeeds


rogeralsing:
TY!

Issue #50, opened 2016-12-20
C# PID.Stop() fails with an "Unknown host" exception


r1sc:
Trying to call pid.Stop() fails on local actors with an "Unknown host" exception.
I managed to "fix" it by modifying TryAdd in ProcessRegistry.cs:

```csharp
public ValueTuple<PID, bool> TryAdd(string id, ActorRef aref)
{
	var pid = new PID()
	{
		Id = id,
		Ref = aref, //cache aref lookup
		Host = "nonhost" // local               <---- This fixes the error
	};
	var ok = _localActorRefs.TryAdd(pid.Id, aref);
	return ValueTuple.Create(pid, ok);
}
```

Maybe I am missing something here?

rogeralsing:
Your fix is correct, the C# support is very much Alpha so this bug surfaced now that Remoting was added.
PR? ;)

r1sc:
Fixed in #52

PR #51, opened 2016-12-20
C# Fixes the "Unknown host" error for local actors

r1sc:
As discussed in #50


r1sc:
It would probably be nicer to use ProcessRegistry's Host-property as implemented in the Go-version. I'm going to close this PR and fix that.

PR #70, opened 2017-01-13
EscalateDirective causes a nil pointer deference when applied to root actor

bryanpkc:
If the supervision strategy for a root-level actor returns EscalateDirective, the program would crash. Instead, the actor should be stopped without bringing down the other actors in the program.


rogeralsing:
Thanks @bryanpkc! 
We need a signed CLA before merging.

PR #75, opened 2017-01-19
Fixed ConsistentHashRouter and modified router actor

samanbarghi:
Fixed concurrency issues with consistenthashrouter, and added a test to avoid the same problem in the future. Also, modified the router actor to stop the removed node after it was removed from the local data structures.


rogeralsing:
Looks good to me, replace the stop delay with a poison pill and we are good to go.

samanbarghi:
Done, but do you mean replacing the delay as well or just the Stop method call?

PR #81, opened 2017-01-24
Partitioned PID Cache Actors

rogeralsing:
This PR introduces a new way to resolve cached cluster PID's.
Instead of having a global map of name -> PID.
We now hide all this inside a dedicated actor for this.
We also run this actor in multiple instances behind a consistent has pool.
This means that reading and writing to the cache will have less contention as we can have as many partitions as we'd like.

### DIFF
--- a/actor/future.go
+++ b/actor/future.go
@@ -46,7 +46,26 @@ func (f *Future) PID() *PID {
 
 // PipeTo forwards the result or error of the future to the specified pids
 func (f *Future) PipeTo(pids ...*PID) {
+	f.cond.L.Lock()
 	f.pipes = append(f.pipes, pids...)
+	//for an already completed future, force push the result to targets
+	if f.done {
+		f.tell()
+	}
+	f.cond.L.Unlock()
+}
+
+func (f *Future) tell() {
+	var m interface{}
+	if f.err != nil {
+		m = f.err
+	} else {
+		m = f.result
+	}
+	for _, pid := range f.pipes {
+		pid.Tell(m)
+	}
+	f.pipes = nil
 }
 
 func (f *Future) sendToPipes() {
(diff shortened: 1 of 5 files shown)


remote/remote_process.go
```diff
@@ -27,7 +27,7 @@ func sendRemoteMessage(pid *actor.PID, message interface{}, sender *actor.PID) {
 		envelope, _ := serialize(msg, pid, sender)
 		endpointManagerPID.Tell(envelope)
 	default:
-		logerr.Printf("failed, trying to send non Proto %s message to %v", reflect.TypeOf(msg), pid)
+		logerr.Printf("failed, trying to send non Proto %s message to %v from %v", reflect.TypeOf(msg), pid, sender)
```
rogeralsing: Added this to see who sent the broken message (a result of the # name problem)

actor/future.go
```diff
@@ -46,27 +46,37 @@ func (f *Future) PID() *PID {
...
-	if f.pipes == nil {
-		return
+	//for an already completed future, force push the result to targets
+	if f.done {
+		f.tell()
 	}
+	f.cond.L.Unlock()
+}
 
+func (f *Future) tell() {
```
rogeralsing: this is a new internal method for Futures to send to all the recipients.
This is called both when the Future completes, and if anyone calls PipeTo after the Future is done.

cluster/pid_cache.go
```diff
@@ -1,29 +1,116 @@
...
+	name      string
+	kind      string
+	respondTo *actor.PID
 }
 
-func (c *pidCache) Add(key string, pid *actor.PID) {
-	c.lock.Lock()
-	c.Cache[key] = pid
-	c.lock.Unlock()
+type pidCacheRequest struct {
```
rogeralsing: Similar to remote.ActorPidRequest, but tailored to work with the consistent hashrouter over the name field.

PR #88, opened 2017-02-02
Fix unused assignment

artemnikitin:
Fix unused assignment

* Initially, you creating variable `name` and assigning to her value of `msg.Name` and checks for an empty name. 
* But in the end `msg.Name` using to create an actor `actor.SpawnNamed(&props, "Remote$"+msg.Name)` where it make sense to use `name`


rogeralsing:
Good catch! thanks

Issue #89, opened 2017-02-05
Use govendor to manage dependencies


jolestar:
The description of dependencies in readme and updateproto.bat is outdated，and building project is hard for newbie。

So, recommended use [govendor](https://github.com/kardianos/govendor) to manage dependency.

dahankzter:
Why not `dep`? https://github.com/golang/dep

I realize I am late to the game on this one but it would be cool to support the would be standard.

jolestar:
@dahankzter I try go dep, but found it not mature enough.
Of course, using any kind of dependency management tool is better than nothing.

dahankzter:
I don't know about the maturity. In the last gophercon Sam Boyer seems to say it is ready https://www.youtube.com/watch?v=5LtMb090AZI&index=17&list=PL2ntRZ1ySWBdD9bru6IR-_WXUgJqvrtx9 
It doesn't mean that it works her but it is approaching readiness fast.

ansrivas:
@stuartcarnie, @rogeralsing 
Wondering if there are any plans to migrate to `dep`. Apparently now that the project has attained production-level readiness as mentioned [here](https://github.com/golang/dep#current-status) ?

PR #90, opened 2017-02-05
Use govendor to manage dependencies

jolestar:
https://github.com/AsynkronIT/protoactor-go/issues/89


stuartcarnie:
Thanks for the pull request. We are very interested in locking dependencies to specific versions in order to ensure repeatable builds. [dep](https://github.com/golang/dep) is emerging as the standard for vendoring dependencies in Go, and therefore we are following this project closely in anticipation of its availability. Once generally available, we'll consider it as the official tool to vendoring.

jolestar:
Ok。But govendor just use go's vendor dir to manage dependency, it should be compatible with the official dep tools. So, It can be a easy and temporary solution for dependency now.

PR #95, opened 2017-02-13
Separated dispatcher from the mailbox

samanbarghi:
Moved checking throughput and yielding functionalities to dispatcher
code. This helps with separation of concepts and provide a cleaner
way to add new dispatchers. Functions such as `AfterStart` and
`BeforeTermiante` were added to improve the functionality further. 
This can add a bit of overhead.

### DIFF
--- a/mailbox/dispatcher.go
+++ b/mailbox/dispatcher.go
@@ -1,20 +1,45 @@
 package mailbox
 
+import "runtime"
+
 type Dispatcher interface {
 	Schedule(fn func())
 	Throughput() int
+	AfterStart()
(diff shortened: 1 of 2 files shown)


rogeralsing:
IMO this kind of changes needs to go in the issue tracker first for discussion.
While I do see the benefits of this, this would also require each actor to have their own unique Dispatcher to keep track of the index. 
Also, there already is the MailboxStatistics which do provide similar interception points.

samanbarghi:
Sure. let me create an issue and point the changes.

PR #108, opened 2017-03-18
Ensure a user message is sent only once when outbound middleware is used

### LINKED ISSUE #107: Duplicated messages when using outbound middleware
When a local context has outbound middleware, every message is sent to the receiver more than once:
- Using `Tell` or `Request`: twice
- Using `RequestFuture`: three times

I'll submit a bugfix.

meAmidos:
This fixes #107


rogeralsing:
Good catch!

Issue #105, opened 2017-03-15
Middleware and localContextReceiver


meAmidos:
Currently, a middleware chain is constructed in such [a way](https://github.com/AsynkronIT/protoactor-go/blob/4f4e1d1ccb8d497180e2ed203852781813b66d26/actor/props.go#L49) that the final element of the chain is the `localContextReceiver` which invokes the actual actor's `Receive()` method. The problem is that localContextReceiver expects the context it gets as an argument to be of the `localContext` type:

```
func localContextReceiver(ctx Context) {
	a := ctx.(*localContext)
	if _, ok := a.message.(*PoisonPill); ok {
		a.self.Stop()
	} else {
		a.receive(ctx)
	}
}
```

So, all the way through the chain the context is propagated as an interface type (`actor.Context`) and only in the end, it is asserted to a specific type. If the `localContextReceiver` could avoid this assertion and pass through the context simply as `actor.Context` interface type, this would make way for a whole new category of middleware which would be able to swap the context to something custom. Now, such swapping works only until the context reaches the point of assertion: `a := ctx.(*localContext)`.

For example, imagine a testing middleware where a custom context object would intercept and register all the calls for later examination in tests.

A brief code research shows that the idea is most likely viable, but would require some refactoring. Would you consider a pull request for this?

Alternatively, is there any other way to achieve the same level of interception in a middleware?

rogeralsing:
Absolutely interested. but please check the different perf benchmarks so any change do not affect perf in a significant way.
cc @stuartcarnie

meAmidos:
So, this issue turned out to be a story with a happy ending. Merging of the PR unblocked the possibility of message interception, and I was able to use this mechanic to build a testing middleware  - exactly as it was planned in the beginning. I hope, this could be helpful for someone as an example of middleware or as a tool for writing tests: https://github.com/meAmidos/gopactor.

rogeralsing:
@meAmidos this looks amazing, I'm going to have to dig deeper into this and see if we can replicate the same features in C#

PR #109, opened 2017-04-03
Middleware refactoring

meAmidos:
This addresses the issue #105.

Inbound and outbound middleware chains are now built not in props, but at the moment of a local context creation. It allows capturing the local context in a closure and using it in a chain. The immediate benefit is that a context interface value can now be passed through the inbound middleware chain till the very end without asserting it to any particular type. Any middleware can now substitute the context value and pass it to the next element of the chain.

A few panicking tests have been fixed along the way.

### DIFF
--- a/actor/future_test.go
+++ b/actor/future_test.go
@@ -5,7 +5,6 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/mock"
 )
 
 func TestFuture_PipeTo_Message(t *testing.T) {
@@ -33,7 +32,9 @@ func TestFuture_PipeTo_Message(t *testing.T) {
 	fp, _ := ref.(*futureProcess)
 
 	fp.SendUserMessage(f.pid, "hello", nil)
-	mock.AssertExpectationsForObjects(t, p1, p2, p3)
+	p1.AssertExpectations(t)
+	p2.AssertExpectations(t)
+	p3.AssertExpectations(t)
 	assert.Empty(t, fp.pipes, "pipes were not cleared")
 }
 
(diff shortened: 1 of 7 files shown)


rogeralsing:
Thanks!
Looking good from a conceptual perspectie. 
But I need to have a look at how this performs in terms of spawn weight.
Doing the middleware chain construction this way will add both CPU and memory pressure.

meAmidos:
I've added a simple benchmark for spawning an actor. The results vary from run to run and mostly fit into the range 4500-7000 ns/op with no consistent difference between versions before and after refactoring. Here are some concrete results:

## Before
```
BenchmarkLocalContext_SpawnWithMiddleware1-4  200000 6656 ns/op
BenchmarkLocalContext_SpawnWithMiddleware2-4  200000 7306 ns/op
BenchmarkLocalContext_SpawnWithMiddleware5-4  300000 5405 ns/op
```

## After
```
BenchmarkLocalContext_SpawnWithMiddleware0-4  200000 5845 ns/op
BenchmarkLocalContext_SpawnWithMiddleware1-4  200000 6545 ns/op
BenchmarkLocalContext_SpawnWithMiddleware2-4  200000 6808 ns/op
BenchmarkLocalContext_SpawnWithMiddleware5-4  300000 5942 ns/op
```

So, basically, there is no any noticeable performance regression. I have not measured CPU and memory usage, though.

I can see an alternative way of achieving the main goal of this PR without creating a middleware chain on every spawn. However, it would require changing the signature of inbound middleware functions which is a breaking change. I am not sure if you are open to such changes.

By the way, if you _are_ open to considering breaking changes, I would rather use this possibility to ~break something else~ to revisit the whole middleware/plugin approach. But this is a big story for another PR :)

rogeralsing:
We are absolutely open for discussions on how the middleware should work.
There are a few things we would like to accomplish having the middleware.

1) Testability, it makes life easier when you can test what goes in, or out of each actor.
2) Tracing, e.g. allowing us to plug in Zipkin to the actors and trace calls between them.
*) All this while still being extremely lightweight

The perf numbers looks good here.

meAmidos:
It's great to know the project is open to changes. Testability is exactly the reason why I started to dig into the middleware implementation.

Is there anything I could do to facilitate the merging of this PR?

rogeralsing:
Merged, thank you for the contribution!
Please join in at the gitter chat https://gitter.im/AsynkronIT/protoactor
As we try to discuss design topics there and it would be great if we can align this into .NET too.

PR #111, opened 2017-04-11
Update README for pid.Stop()

johannesh:


johannesh:
Specifically when a call to Spawn() is immediately followed by Stop() or GracefulStop() I get panics. A sleep statement helps in this case, so I suspect a race condition.

Example: https://gist.github.com/johannesh/0d5805a45d7454b52ef5076db2a500b9

Happy to provide more details

johannesh:
:+1:

PR #112, opened 2017-04-21
Fix a bug in AwaitFuture

morikuni:
After the Receive function returns, ctx.message become nil.
So, the ctx.message have to be bound to a local variable.


morikuni:
This fixes #114

rogeralsing:
:+1:

PR #137, opened 2017-07-07
fix roundrobin_router when access the maximum of int32 will produce a…

qgymje:
… minus index value which will cause panic

### DIFF
--- a/router/roundrobin_router.go
+++ b/router/roundrobin_router.go
@@ -52,6 +52,10 @@ func (config *roundRobinGroupRouter) CreateRouterState() Interface {
 
 func roundRobinRoutee(index *int32, routees []actor.PID) actor.PID {
 	i := int(atomic.AddInt32(index, 1))
+	if i < 0 {
+		*index = 0
+		i = 0
+	}
 	mod := len(routees)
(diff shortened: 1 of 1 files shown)


qgymje:
maybe "let it crash" ?

alex023:
The method [roundRobinRoutee]just used by [RouteMessage],and input param ```index *int32```is only ```&state.index``` which default minimum value is 0. 
Because unique creation method is [[source code]](https://github.com/AsynkronIT/protoactor-go/blob/dev/router/roundrobin_router.go#L45-#L47)：
```golang
func (config *roundRobinPoolRouter) CreateRouterState() Interface {
	return &roundRobinState{}
}
```
So，i think this modification is unnecessary at present .

cpx86:
Nice catch!

@alex023 The `index` field is incremented each time a message is sent. So if `index` is int32 max, `i` will overflow to a negative (int32 min), which after modulus becomes a negative index in the array access :)

qgymje:
@alex023 It's not about the minimum value of state.index, it's about the maximum value of state.index

alex023:
Oh, I did not think about that。v_v

Issue #140, opened 2017-07-19
persistence: uninitialized Mixin.providerState and infinite recursive message reception


RussellLuo:
## Example code

To try the persistence package, I write a simple example as follows:

```go
package main

import (
	"fmt"

	"github.com/AsynkronIT/goconsole"
	"github.com/AsynkronIT/protoactor-go/actor"
	"github.com/AsynkronIT/protoactor-go/persistence"
	"github.com/golang/protobuf/proto"
)

type ProviderState struct {
	events []proto.Message
}

func (s *ProviderState) Restart() {
}

func (s *ProviderState) GetSnapshotInterval() int {
	return -1 // snapshotting is disabled
}

func (s *ProviderState) GetSnapshot(actorName string) (snapshot interface{}, eventIndex int, ok bool) {
	return nil, 0, false
}

func (s *ProviderState) GetEvents(actorName string, eventIndexStart int, callback func(e interface{})) {
	for _, e := range s.events[eventIndexStart:] {
		callback(e)
	}
}

func (s *ProviderState) PersistEvent(actorName string, eventIndex int, event proto.Message) {
	s.events = append(s.events, event)
}

func (s *ProviderState) PersistSnapshot(actorName string, eventIndex int, snapshot proto.Message) {
}

type Provider struct{}

func (p *Provider) GetState() persistence.ProviderState {
	return &ProviderState{events: []proto.Message{}}
}

type protoMsg struct{}

func (p *protoMsg) Reset()         {}
func (p *protoMsg) String() string { return "" }
func (p *protoMsg) ProtoMessage()  {}

type Actor struct {
	persistence.Mixin
}

type Message struct {
	protoMsg
}

type Snapshot struct {
	protoMsg
}

func (a *Actor) Receive(ctx actor.Context) {
	switch msg := ctx.Message().(type) {
	case *actor.Started:
		fmt.Println("started")
	case *Message:
		fmt.Printf("received message %+v\n", msg)
		a.PersistReceive(msg)
	case *persistence.RequestSnapshot:
		fmt.Println("request to snapshot")
		a.PersistSnapshot(&Snapshot{})
	case *Snapshot:
		fmt.Println("recover internal state from snapshot")
	case *persistence.ReplayComplete:
		fmt.Println("all events are replayed")
	}
}

func main() {
	props := actor.FromInstance(&Actor{}).WithMiddleware(persistence.Using(&Provider{}))
	pid := actor.Spawn(props)
[...]

PR #141, opened 2017-07-19
Fix persistence issue #140

RussellLuo:
Fixed  issue #140.


rogeralsing:
This looks good. the Started message was initially processed synchronously and that was the reason why we had the replay command.

PR #144, opened 2017-07-20
Fix log levels

Anber:
Debug, Info and Error methods have published events with the same log level.


rogeralsing:
Thanks for spotting this!

PR #150, opened 2017-07-26
Update README.md

cemremengu:
Changed `StartServer` to `Start` since it's undefined (I think the API changed)


cpx86:
Nice catch, thanks!

PR #151, opened 2017-07-26
Update README.md

cemremengu:
I think the API changed for this one also

`actor.ProcessRegistry.Register("myactor", pid)` does not work anymore


cemremengu:
There is `actor.ProcessRegistry.Add` but it needs a `process` and wont accept `pid` so not sure

Issue #154, opened 2017-07-28
In memory persistence provider not go routine safe


dahankzter:
I realize this is probably meant for testing and such but the race detector complained.

It is easy to just wrap a couple of locks and use them.

ethanfrey:
This is resolved, isn't it, can be closed?

I'm just looking over open issues, to get a feel for the state. It seems many = of the open ones are questions or have fixes, I think a bit of cleanup would make the github look a bit more inviting to new users.

PR #155, opened 2017-07-28
Making the in memory persistence provider goroutine safe

dahankzter:
I have a pull request #154 adding a couple of locks one for snapshots and one for the events map. It is not optimal from a performance perspective but should be safe.

### DIFF
--- a/persistence/in_memory_provider.go
+++ b/persistence/in_memory_provider.go
@@ -1,6 +1,10 @@
 package persistence
 
-import "github.com/golang/protobuf/proto"
+import (
+	"sync"
+
+	"github.com/golang/protobuf/proto"
+)
 
 type snapshotEntry struct {
 	eventIndex int
@@ -9,15 +13,19 @@ type snapshotEntry struct {
 
 type InMemoryProvider struct {
 	snapshotInterval int
-	snapshots        map[string]*snapshotEntry  // actorName -> a snapshot entry
+	snapshots        map[string]*snapshotEntry // actorName -> a snapshot entry
+	snapshotLock     *sync.Mutex
 	events           map[string][]proto.Message // actorName -> a list of events
+	eventsLock       *sync.Mutex
 }
 
 func NewInMemoryProvider(snapshotInterval int) *InMemoryProvider {
 	return &InMemoryProvider{
 		snapshotInterval: snapshotInterval,
 		snapshots:        make(map[string]*snapshotEntry),
+		snapshotLock:     &sync.Mutex{},
 		events:           make(map[string][]proto.Message),
+		eventsLock:       &sync.Mutex{},
 	}
 }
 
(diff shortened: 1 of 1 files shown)


RussellLuo:
@dahankzter Firstly, thanks a lot for pointing out the concurrency-safety problem, and the changes you submitted look feasible.

But as far as I can see, while taking concurrency-safety into consideration, there are more possible improvements to make. I'm now planning to refactor the in-memory persistence provider as follows:

1. Merge `snapshots` map and `events` map into one, so we can just use one mutex
2. Use `sync.RWMutex` instead of `sync.Mutex` to get better read performance
3. Since each actor is working in a standalone goroutine, snapshotting and event-persistence will happen in sequence, so we can just lock the loading/initialing operation of map entry (for each actor), thus getting better performance while still be concurrency-safe

Any more suggestions are appreciated, thanks again!

RussellLuo:
@dahankzter Here's the new pull request #156, it would be nice if you can take a review 😄

dahankzter:
This was along the lines I was considering first but opted for a very simple option since I didn't know if this provider was to actually be used.

If the event storing is always in the same goroutine then that should be safe as well. Some tests should maybe be added.

If there is no requirements for a particular Go version then for 1.9 we could perhaps use the new sync.Map.

Will the contributor status update automatically or do I have to push something new?

RussellLuo:
> This was along the lines I was considering first but opted for a very simple option since I didn't know if this provider was to actually be used.

Yes I saw it :-) The in-memory persistence provider is mainly for trivial scenario and testing purpose, but at the same time, it's also a reference implementation maintained in the core library. So I think it's also important to keep its efficiency and safety, as well as simplicity.

> If the event storing is always in the same goroutine then that should be safe as well. Some tests should maybe be added.

The event/snapshot storing is only concurrency-safe in one actor, there are still safety issues if multiple actors use the same provider concurrently.

> If there is no requirements for a particular Go version then for 1.9 we could perhaps use the new sync.Map.

I think the `sync.RWMutex` is also efficient enough since "map write operation" rarely happens in actor model. That being said, `sync.Map` is awesome if we upgrade to Go 1.9, and then we only need to change [loadOrInit](https://github.com/RussellLuo/protoactor-go/blob/refactor-in-memory-persistence-provider/persistence/in_memory_provider.go#L31) as follows:

```go
func (provider *InMemoryProvider) loadOrInit(actorName string) (e *entry, loaded bool) {
	e, loaded := provider.store.LoadOrStore(actorName, &entry{})
	return e.(*entry), loaded
}
```

> Will the contributor status update automatically or do I have to push something new?

This library still lacks docs, tests, etc, and it may still have some issues. We're looking forward to more contributions from you!

rogeralsing:
Can I close this as #156 has been merged.

Issue #157, opened 2017-07-30
router: issues of the method process.SendUserMessage


RussellLuo:
Given the following testing code (modified from [consistent_hash_router_test.go](https://github.com/AsynkronIT/protoactor-go/blob/dev/router/consistent_hash_router_test.go)):

```go
package main

import (
	"strconv"

	"github.com/AsynkronIT/goconsole"
	actor "github.com/AsynkronIT/protoactor-go/actor"
	"github.com/AsynkronIT/protoactor-go/router"
)

type myMessage struct {
	i   int
	pid *actor.PID
}
type getRoutees struct {
	pid *actor.PID
}

func (m *myMessage) Hash() string {
	return strconv.Itoa(m.i)
}

type routeeActor struct{}
type managerActor struct {
	set  []*actor.PID
	rpid *actor.PID
}

func (state *routeeActor) Receive(context actor.Context) {
	switch msg := context.Message().(type) {
	case *myMessage:
		//log.Printf("%v got message %d", context.Self(), msg.i)
		msg.i++
	}
}

func (state *managerActor) Receive(context actor.Context) {
	switch msg := context.Message().(type) {
	case *getRoutees:
		state.rpid = msg.pid
		msg.pid.Request(&router.GetRoutees{}, context.Self())
	}
}

func main() {
	rpid := actor.Spawn(router.NewConsistentHashPool(5).WithInstance(&routeeActor{}))

	props := actor.FromInstance(&managerActor{})
	pid := actor.Spawn(props)
	pid.Tell(&getRoutees{rpid})
	console.ReadLine()
}
```

Simply run the above code:

```bash
$ go run main.go
```

and we got an error:

```bash
2017/07/30 17:36:25 [ROUTING] Message must implement router.Hasher &GetRoutees{}
```

which is obviously caused the method [process.SendUserMessage](https://github.com/AsynkronIT/protoactor-go/blob/dev/router/process.go#L20-L29):

```go
func (ref *process) SendUserMessage(pid *actor.PID, message interface{}) {

	msg, sender := actor.UnwrapEnvelope(message)
	// issue 1: We should use `msg` in the following line
	if _, ok := message.(ManagementMessage); ok {
		r, _ := actor.ProcessRegistry.Get(ref.router)
		r.SendUserMessage(pid, msg)
	} else {
		ref.state.RouteMessage(msg, sender)
	}
}
```

After this issue is addressed, another one occurs:

```bash
[...]

PR #158, opened 2017-07-30
Always send the original (possibly enveloped) message to router actor

RussellLuo:
Fixed issue #157.

BTW: I swapped the order of "if ... else ..." for better readability (I hope so).

### DIFF
--- a/router/process.go
+++ b/router/process.go
@@ -18,15 +18,17 @@ type process struct {
 }
 
 func (ref *process) SendUserMessage(pid *actor.PID, message interface{}) {
-
 	msg, sender := actor.UnwrapEnvelope(message)
-	if _, ok := message.(ManagementMessage); ok {
-		r, _ := actor.ProcessRegistry.Get(ref.router)
-		r.SendUserMessage(pid, msg)
-	} else {
+	if _, ok := msg.(ManagementMessage); !ok {
 		ref.state.RouteMessage(msg, sender)
+	} else {
+		r, _ := actor.ProcessRegistry.Get(ref.router)
+		// Always send the original message to the router actor,
+		// since if the message is enveloped, the sender need to get a response.
+		r.SendUserMessage(pid, message)
 	}
 }
+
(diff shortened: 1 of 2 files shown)


rogeralsing:
Nice catch!

Btw. in the refactored Router implementation in Kotlin, the RouterProcess owns the RouterActor.
The current Go implementation has the RouterActor as a separate LocalProcecss.
This causes problems if you e.g. Watch the router.

Currently: 
RouterProcess has its own PID
RouterActor has its own PID

The RouterProcess PID is what you get returned from the newXXXRouter.
if you Watch the RouterProcess PID, and the router shuts down, it will respond with the PID from the RouterActor.

This also puts unneeded pressure on the process registry as there are two entries for each router.

So we should aim to copy the Kotlin/C# behavior to fix this

RussellLuo:
> The RouterProcess PID is what you get returned from the newXXXRouter.
if you Watch the RouterProcess PID, and the router shuts down, it will respond with the PID from the RouterActor.

Sorry, I didn't catch it. As far as I can see in the [current implementation](https://github.com/AsynkronIT/protoactor-go/blob/dev/router/process.go#L32-L65), there is no logic that RouterProcess delegates Watch/Unwatch message to RouterActor's LocalProcess:

```go
...
case *actor.Watch:
	if atomic.LoadInt32(&ref.stopping) == 1 {
		if r, ok := actor.ProcessRegistry.Get(msg.Watcher); ok {
			r.SendSystemMessage(msg.Watcher, &actor.Terminated{Who: pid})
		}
		return
	}
	ref.mu.Lock()
	ref.watchers.Add(msg.Watcher)
	ref.mu.Unlock()

case *actor.Unwatch:
	ref.mu.Lock()
	ref.watchers.Remove(msg.Watcher)
	ref.mu.Unlock()
...
```

If the RouterActor shuts down by calling `context.Self().Stop()`, then Stop message will be sent to the RouterActor's LocalProcess, but maybe nothing will happen because no watchers exists there?

> This also puts unneeded pressure on the process registry as there are two entries for each router.

Yes, it is.

Issue #159, opened 2017-08-01
Access to remote actor should be added to the decision


alex023:
when access remote actor with bad address，will  keep the error, resulting in a large system overhead.

**client code**
```go
package main

import (
	"fmt"
	"github.com/AsynkronIT/goconsole"
	"github.com/AsynkronIT/protoactor-go/actor"
	"github.com/AsynkronIT/protoactor-go/remote"
	"time"
)

func main() {
	remote.Start("127.0.0.1:9999")
	//if server is not started or not exist
	server := actor.NewPID("127.0.0.1:1111", "actorName")
	fmt.Println(server)
	_, err := server.RequestFuture(&struct{}{}, time.Second).Result()
	if err != nil {
		fmt.Println(err)
	}

	console.ReadLine()
}
```
Once the client is run, it will be given an unlimited number of errors,even if  the remote actor is started.
```
2017/08/01 08:24:39 [REMOTE] Started EndpointWriter address="127.0.0.1:1111" 
2017/08/01 08:24:39 [REMOTE] EndpointWatcher connecting address="127.0.0.1:1111" 
2017/08/01 08:24:39 [REMOTE] [ACTOR] Recovering actor="127.0.0.1:9999/$1/$3" reason="runtime error: invalid memory address or nil pointer dereference" stacktrace="vendor/github.com/AsynkronIT/protoactor-go/remote.(*endpointWriter).initializeInternal:46" 
2017/08/01 08:24:39 [REMOTE] Started EndpointWriter address="127.0.0.1:1111" 
2017/08/01 08:24:39 [REMOTE] EndpointWatcher connecting address="127.0.0.1:1111" 
2017/08/01 08:24:39 [REMOTE] [ACTOR] Recovering actor="127.0.0.1:9999/$1/$3" reason="runtime error: invalid memory address or nil pointer dereference" stacktrace="vendor/github.com/AsynkronIT/protoactor-go/remote.(*endpointWriter).initializeInternal:46" 
2017/08/01 08:24:39 [REMOTE] Started EndpointWriter address="127.0.0.1:1111" 
2017/08/01 08:24:39 [REMOTE] EndpointWatcher connecting address="127.0.0.1:1111" 
...
```
Whether it should join a mechanism to control system overhead？

RussellLuo:
[...]

PR #160, opened 2017-08-01
Handle gRPC connection error properly

RussellLuo:
Fixed issue #159.

### DIFF
--- a/remote/endpoint_writer.go
+++ b/remote/endpoint_writer.go
@@ -43,6 +43,9 @@ func (state *endpointWriter) initializeInternal() error {
 	state.conn = conn
 	c := NewRemotingClient(conn)
 	resp, err := c.Connect(context.Background(), &ConnectRequest{})
+	if err != nil {
+		return err
+	}
 	state.defaultSerializerId = resp.DefaultSerializerId
 
 	//	log.Printf("Getting stream from address %v", state.address)


rogeralsing:
There are things we should consider here too.
If the Connect fails for whatever reason, the EndpointWriter will be in an invalid state.
It will be restarted if anyone tries to send something via it.
But maybe there are some design ideas we can get out of it.
e.g should the EndpointWriter change state into Invalid?
Should it stop itself / Ask the EndpointManager to remove it from it's internal routing table?

Thoughts?

RussellLuo:
After taking a quick look at the related code, I think maybe we need:
1. send a message, much similar to the message `EndpointTerminatedEvent` or something new named `EndpointUnavailableEvent`, to indicate the endpointWriter is broken
2. then, stop the endpointWriter itself

In order to get the above steps to work, maybe we should also change the [`EndpointTerminatedEvent ` handler in endpointWriter](https://github.com/AsynkronIT/protoactor-go/blob/dev/remote/endpoint_manager.go#L53-L56) to remove broken endpoint from the internal `state.connections`, after which a new endpoint can be spawned when necessary.

alex023:
@RussellLuo `EndpointUnavailableEvent` is good choice

rogeralsing:
I will merge this as it fixes the direct problem at hand, and we can discuss a more complete solution as a separate issue

PR #174, opened 2017-09-14
Fix partition naming rule

PotterDai:
Old partition naming rule adds "#" prefix.
This is different with [protoactor-dotnet](https://github.com/AsynkronIT/protoactor-dotnet/blob/13bf29369764e8aa3716a5c3246467ab951794ea/src/Proto.Cluster/Partition.cs#L22), making them cannot talk to each other.


rogeralsing:
👍

PR #175, opened 2017-09-14
fixed #172

### LINKED ISSUE #172: remote.Watch can't support one local actor monitor many actor in the same remote endpoint
We are using proctor in our business.But in some complex scene , there are some problem in remote.Watch

## Problem Description:
 We can provide a simplified large-scale chat room to simulate the scene of the problem。Many users share in topics through serval proxy servers in the chat server . We must support：
- Feature1：Normally, notify a user published message to other online users.
- Feature2：restart Chat Server，Proxy Server Actor can monitor ,and reconnect after Chat Server restarted 。
- Feature3：One Proxy Server shutdown，Topic[Actor]can monitor and refresh “online member list”.
- Feature4：User disconnect,Topic[Actor]can monitor, and move it from “online member list”。

The flow picture show the topology we want:
image
 Topic is Actor, we use `Watch`to monitor whether the Agent[Actor] is alive, the code just like `examples/chat/server`：
```go
func notifyAll(clients *hashset.Set, message interface{}) {
	fmt.Printf("send msg %v %d client\n",message,clients.Size())
	for _, tmp := range clients.Values() {
		client := tmp.(*actor.PID)
		client.Tell(message)
	}
}

func main() {
	remote.Start("127.0.0.1:8080")
	clients := hashset.New()
	props := actor.FromFunc(func(context actor.Context) {
		switch msg := context.Message().(type) {
		case *actor.Started:
			println("room is started!")
		case *actor.Stopping:
			println("room is stopping!")

		case *messages.Connect:
			clients.Add(msg.Sender)
			msg.Sender.Tell(&messages.Connected{Message: "Welcome!"})
			fmt.Println("server Watch client actor")
			context.Watch(msg.Sender)
		case *actor.Terminated:
			//want remove it from online member list,when the agent[/the node] is disconnect
			clients.Remove(msg.Who)
		case *messages.SayRequest:
			notifyAll(clients, &messages.SayResponse{
				UserName: msg.UserName,
				Message:  msg.Message,
			})
		case *messages.NickRequest:
			notifyAll(clients, &messages.NickResponse{
				OldUserName: msg.OldUserName,
				NewUserName: msg.NewUserName,
			})
		}
	})
	remote.Register("room",props)
	console.ReadLine()
}
```
### For Feature3:
When the proxy server shutdown, the Chat Server still try to send the message to the Agent[Actor] in the proxy server.It can not monitor some agent[Actor] are terminated.
image
In endpoint_watcher.go,the struct endpointWatcher just store only one remote actor.PID.
```
type endpointWatcher struct {
	address string          //<—Same the endpoint
	watched map[string]*actor.PID //<—one actor just watch only one remote actor
}
```
when a new remote actor was watch by local actor, the watched will reset to a new actor PID.and the older PID is throw away。
And now, the proxy 1  shutdown, the endpointWatcher just notify the newer remote actor is terminated.
And when this proxy server restart, the client can not connect the Chat Server ,because the endpointWatcher call ctx.SetBehavior(state.Terminated).
 
### For Feature4:
When the actor is disconnect, the Chat Server Topic can not determine.
image
endpointWatcher can receive remoteTerminate event, but the watchee ptr is not the same as the previous cache in watched.
The source code:
```
case *remoteTerminate:
	delete(state.watched, msg.Watcher.Id)
	terminated := &actor.Terminated{
				Who:               msg.Watchee,    //<—msg.Watchee ptr is not same as  localcontext.watching in local_context.go 
				AddressTerminated: false,
			}
	ref, ok := actor.ProcessRegistry.GetLocal(msg.Watcher.Id)
	if ok {
		ref.SendSystemMessage(msg.Watcher, terminated)
	}
```
So, is connect sequence is A、B、C, the  actor of A ptr is lost.

## Suggest:
If we have no problem with the understanding of the Watch, I think remote_watcher.go should be refactored. 
1. endpointWatcher.watched should is ”map[string]map[string]*actor.PID”.

just like: ```
type endpointWatcher struct {
	address string
	watched map[string]*watchedPIDs //key is the watching PID string, value is the watched PIDs
}

type watchedPIDs struct {
	pids map[string]*actor.PID  
}

func (wps *watchedPIDs) add(pid *actor.PID) {
	if wps.pids == nil {
		wps.pids = make(map[string]*actor.PID)
	}
	wps.pids[pid.Id] = pid
}
func (wps *watchedPIDs)get(id string)(*actor.PID,bool){
	pid,founded:=wps.pids[id]
	return pid,founded
}
func (wps *watchedPIDs) remove(pid *actor.PID) {
	delete(wps.pids, pid.Id)
	if len(wps.pids) == 0 {
		wps.pids = nil
	}
}
func (wps *watchedPIDs) all() []*actor.PID {
	var result []*actor.PID
	for _, pid := range wps.pids {
		result = append(result, pid)
	}
	return result
}
func (wps *watchedPIDs) clean() {
	wps.pids = nil
}
func (wps *watchedPIDs) size() int {
	if wps.pids == nil {
		return 0
	} else {
		return len(wps.pids)
	}

}

```
2. do not use ctx.SetBehavior(state.Terminated)
after call this function, it can not Watch again when  the remote server restarted.

## Finally:
We continue to debug further to avoid missing.
I want to know if there is an error in use in the above scenario， please give some comments.
If ok, we can pull the request about an week later.

### PR DESCRIPTION
**Pull request by alex023**
there still more jobs todo:
1. when remote endpoint is shutdown, can not use Terminate State[/Behavior]
2. need add more events for control endpoint_watcher behavior change .

### DIFF
--- a/remote/endpoint_watcher.go
+++ b/remote/endpoint_watcher.go
@@ -34,8 +34,8 @@ func (state *endpointWatcher) Receive(ctx actor.Context) {
 
 		if watchedPIDs, founded := state.watched[msg.Watcher.Id]; founded {
 			//The cache object pointer is different from the temporary object pointer interpreted by the reflector
-			if cacheWatchee, deleted := watchedPIDs.Remove(msg.Watchee.Id); deleted {
-				watchee = cacheWatchee
+			if cachedWatchee, deleted := watchedPIDs.Remove(msg.Watchee.Id); deleted {
+				watchee = cachedWatchee
 
 				if watchedPIDs.Size() == 0 {
 					delete(state.watched, msg.Watcher.Id)
@@ -53,11 +53,11 @@ func (state *endpointWatcher) Receive(ctx actor.Context) {
 
 	case *EndpointTerminatedEvent:
 		plog.Info("EndpointWatcher %v  handling terminated", log.String("address", state.address), log.String("address", state.address))
-		for id, pids := range state.watched {
+		for id, watchedPIDs := range state.watched {
 			//try to find the watcher ID in the local actor registry
 			localWatcher, ok := actor.ProcessRegistry.GetLocal(id)
 			if ok {
-				for _, watchee := range pids.All() {
+				for _, watchee := range watchedPIDs.All() {
 					//create a terminated event for the Watched actor
 
 					terminated := &actor.Terminated{
@@ -68,7 +68,7 @@ func (state *endpointWatcher) Receive(ctx actor.Context) {
 					//send the address Terminated event to the Watcher
 					localWatcher.SendSystemMessage(watcher, terminated)
 				}
-				pids.Clean()
+				watchedPIDs.Clean()
 			}
 		}
 
@@ -80,14 +80,14 @@ func (state *endpointWatcher) Receive(ctx actor.Context) {
 
 	case *remoteWatch:
 		var (
-			pids    *PIDSet
-			founded bool
+			watchedPIDs *PIDSet
+			founded     bool
 		)
-		if pids, founded = state.watched[msg.Watcher.Id]; !founded {
-			pids = &PIDSet{}
-			state.watched[msg.Watcher.Id] = pids
+		if watchedPIDs, founded = state.watched[msg.Watcher.Id]; !founded {
+			watchedPIDs = &PIDSet{}
+			state.watched[msg.Watcher.Id] = watchedPIDs
 		}
-		pids.Add(msg.Watchee)
+		watchedPIDs.Add(msg.Watchee)
 
 		//recreate the Watch command
 		w := &actor.Watch{

### CODE REVIEW

**Comment by rogeralsing** [comment]
I have been thinking that maybe we need to merge endpoint watcher and endpoint writer.
That would allow us to get rid of state/behavior and there would be no need for complex event interactions.

Does this make sense or do you see any problems with this approach?

**Comment by alex023** [comment]
Maybe ,we get rid of state/behavior change ,add Heartbeat Event for FailureDetector is better.

File: remote/endpoint_watcher.go [hunk 1]
```diff
@@ -15,12 +15,12 @@ func newEndpointWatcher(address string) actor.Producer {
 
 type endpointWatcher struct {
 	address string
-	watched map[string]*actor.PID //key is the watching PID string, value is the watched PID
+	watched map[string]*PIDSet //key is the watching PID string, value is the watched PID
```
PotterDai: Probably you can just use *actor.PIDSet, there is already a class dedicated for the PID sets.

alex023: Thank u advise.
thanks for your advice.
I have considered using actor.PID, but in this way, once the business layer uses Watcher.Who, it can not be positioned.Because the actor's address based on protobuf unmarshal is new every time, which is inconsistent with the previous cache of Watcher.Who.
In business layer code:
```golang
//...
	case *messages.Connect:
		log.Printf("Client [%v,%p] connected", msg.Sender,msg.Sender)
		clients.Add(msg.Sender)     //msg.Sender is ptr of actor.PID
		msg.Sender.Tell(&messages.Connected{Message: "Welcome!"})
		context.Watch(msg.Sender) 
	case *actor.Terminated:
			clients.Remove(msg.Who)    //msg.Who ptr is should in clients
			log.Printf("remove after :clients[%+v],who[%p],connected size is %v ", clients,msg.Who,clients.Size())
//...
```
We can catch the object creation in **proto_serializer.go**
```golang
func (protoSerializer) Deserialize(typeName string, bytes []byte) (interface{}, error) {
	protoType := proto.MessageType(typeName)
	if protoType == nil {
		return nil, fmt.Errorf("Unknown message type %v", typeName)
	}
	t := protoType.Elem()    

	intPtr := reflect.New(t)       //<----this is new ptr
	instance := intPtr.Interface().(proto.Message)
	proto.Unmarshal(bytes, instance)

	return instance, nil
}
```
So I have to provide a cache that helps the same actor with the same ptr.

PotterDai: Hey, actually in actor.PIDSet, it is using string as key.
 https://github.com/AsynkronIT/protoactor-go/blob/6c50ad304f4f87567c31f30487343b51aa96c7f3/actor/pidset.go#L5-L8
So it might be OK just using it = )

rogeralsing:
cc @PotterDai is this something we should merge?

PotterDai:
@rogeralsing It should have been fixed already in my previous PR.

PR #180, opened 2017-09-26
Revision for membership

PotterDai:
Since we now have StatusCode. MemberList no longer needs to wait for members to join in.
If no members available, it now just returns Unavailable status code and user can decide retry strategy.
This PR adjusts that.


rogeralsing:
👍 Perfect, this is indeed the correct approach for this problem

PR #182, opened 2017-10-12
Fixes bug in transferring ownership of grains after adding members to a cluster

gviedma:
Preexisting grains do not get transferred to the new node's partition after a node is added to a cluster. Instead, it results in the following stack trace:

```
2017/10/12 11:16:34 [MAILBOX] [ACTOR] Recovering actor="172.18.0.4:8080/activator" reason="spawn: name exists" stacktrace="github.com/AsynkronIT/protoactor-go/remote.(*activator).Receive:125"
2017/10/12 11:16:34 [ACTOR] [SUPERVISION] actor="172.18.0.4:8080/activator" directive="RestartDirective" reason="spawn: name exists"
2017/10/12 11:16:34 [REMOTE] Activator received unknown message message=&{}
```

Also, getMembers should return all members for a partition kind, including those that are not alive. Otherwise, transferring of grains will fail, resulting in multiple activations.

### DIFF
--- a/cluster/memberlist.go
+++ b/cluster/memberlist.go
@@ -11,7 +11,7 @@ import (
 //TODO: this needs to be implemented,we could send a `Request` to the membership actor, but this seems flaky.
 //a threadsafe map would be better
 func getMembers(kind string) []string {
-	res, err := memberlistPID.RequestFuture(&MemberByKindRequest{kind: kind, onlyAlive: true}, 5*time.Second).Result()
+	res, err := memberlistPID.RequestFuture(&MemberByKindRequest{kind: kind, onlyAlive: false}, 5*time.Second).Result()
 	if err == nil {
 		if t, ok := res.(*MemberByKindResponse); ok && len(t.members) > 0 {
 			return t.members
--- a/remote/activator_actor.go
+++ b/remote/activator_actor.go
@@ -122,7 +122,6 @@ func (*activator) Receive(context actor.Context) {
 				StatusCode: ResponseStatusCodePROCESSNAMEALREADYEXIST.ToInt32(),
 			}
 			context.Respond(response)
-			panic(err)
 		} else if aErr, ok := err.(*ActivatorError); ok {
 			response := &ActorPidResponse{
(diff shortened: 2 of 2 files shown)


cluster/memberlist.go
```diff
@@ -11,7 +11,7 @@ import (
 //TODO: this needs to be implemented,we could send a `Request` to the membership actor, but this seems flaky.
 //a threadsafe map would be better
 func getMembers(kind string) []string {
-	res, err := memberlistPID.RequestFuture(&MemberByKindRequest{kind: kind, onlyAlive: true}, 5*time.Second).Result()
+	res, err := memberlistPID.RequestFuture(&MemberByKindRequest{kind: kind, onlyAlive: false}, 5*time.Second).Result()
```
rogeralsing: I'm not sure about this change.
This would mean that nodes that Consul considers to be unhealthy are included in the result.
And when you call `cluster.Get` with a given ID that could belong to an unhealthy node, it would try to communicate with this unhealthy node.

(I'm not sure about (my own?) comment above the code either "reachable and unreachable" )

rogeralsing: I'm thinking if nodes appear as critical when they are not, then this might be an effect of this:

```
func New() (*ConsulProvider, error) {
	client, err := api.NewClient(&api.Config{})
	if err != nil {
		return nil, err
	}
	p := &ConsulProvider{
		client:             client,
		ttl:                3 * time.Second,   
		refreshTTL:         1 * time.Second,
		deregisterCritical: 10 * time.Second,
```

Maybe the ttl and refreshTTL are too narrow and the nodes appear as unhealthy when they are not.
I will investigate

gviedma: Thanks for looking into this @rogeralsing The situation I am running into in my tests is caused by a race condition between a _pidCachePartitionActor_ calling `getMember` to resolve the partition actor's address and _partitionActor_ processing a  `memberJoined` event. In the case of flappy consul health, the `getMember` returns different results between the two and the ownership fails to transfer correctly to the new member.

PR #183, opened 2017-10-12
Revision for membership

PotterDai:
This is same PR with https://github.com/AsynkronIT/protoactor-dotnet/pull/341

This PR contains several revisions and fixes:
1. New Membership structure, cache members by kinds in MemberListActor. This is implemented 
b98d1e8, d63132c
2. When a node gracefully shutdown, its Partition now will transfer all its ownerships not located in itself to other Partitions, similar to what happened when a member join in. This is revised in bb5d790, 36c6785
3. Implemented weighted round robin. This is a feature I personally want a lot, and I think it's very useful for other users as well. Now user can define node weight from 1-10(*see notes), and activator will select which node to activate new actor according to the weight. This is implemented in 18210a1, e5fe500 
4. Add graceful shutdown to the example.  30e974a 

Notes:
The current weighted round robin implementation has a weight limit by 10. This limit will be removed in the future.


rogeralsing:
@PotterDai is this one up to date with the .NET version?

PotterDai:
@rogeralsing Yep, they are the same. So just merge it = )

PR #189, opened 2017-10-16
Grain code generation

rogeralsing:
This PR adds the GrainContext param to codegenerated grains, which is a breaking change.
But the improvement is worth it as grains can now spawn child actors.

System and AutoReceive messages are also filtered from reaching the grain logger

### DIFF
--- a/examples/cluster/shared/protos_protoactor.go
+++ b/examples/cluster/shared/protos_protoactor.go
@@ -10,9 +10,7 @@ import cluster "github.com/AsynkronIT/protoactor-go/cluster"
 
 import proto "github.com/gogo/protobuf/proto"
 import fmt "fmt"
-import (
-	math "math"
-)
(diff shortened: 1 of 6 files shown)


examples/cluster/shared/protos_protoactor.go
```diff
@@ -226,7 +238,11 @@ func (a *HelloActor) Receive(ctx actor.Context) {
 	case *actor.Started:
 		a.inner = xHelloFactory()
 		id := ctx.Self().Id
-		a.inner.Init(id[7:len(id)]) //skip "remote$"
+		a.inner.Init(id[7:]) //skip "remote$"
+
+	case actor.AutoReceiveMessage: //pass
+	case actor.SystemMessage: //pass
```
rogeralsing: Filter out AutoReceive and System messages

PR #190, opened 2017-10-18
Memberlist Get functions can be concurrent

PotterDai:
Same with https://github.com/AsynkronIT/protoactor-dotnet/pull/344


PotterDai:
@rogeralsing Hey Roger, I think this PR should be OK to be merged. It is the same with https://github.com/AsynkronIT/protoactor-dotnet/pull/344 , which is already merged last week.

Also, I think it might be a good time to push a new version to Nuget, since current cluster fixed a bunch of bugs existing in 0.13. What you think?
Thx! = )

PR #192, opened 2017-10-31
Two minor fixes

PotterDai:
- Use go channel instead of mutex in logger.
- Reset timer when receiving timeout callback. Same as https://github.com/AsynkronIT/protoactor-dotnet/pull/346


PotterDai:
@rogeralsing  Hey Roger, those 2 are also small fixes, I'll just go ahead and merge it myself. :)

Issue #202, opened 2017-11-23
The Tell, Request, RequestFuture in actor.Context should be able to pass private envelop.Header


alex023:
## Description
Currently, there is a `Tell`, `Request`,`RequestFuture` method for both **PID** and **Context**, but the latter is able to use Outbox_Middleware. Once we need to pass the message environment variables, we need to have multiple assignments in the middleware implementation. Different Header, Middleware must be different.
I noticed that the `Tell`, `Request`,`RequestFuture` method assign envelop.Header to the global emptyHeader in  **local_context**. If the client application request with header information for debug(just seq or more), this time is difficult to meet. 
Why not pass the head message of the current message by default?

## Suggest
I think it is necessary to pass current header if it is not nil. If we can, we do not need to specify a different middle implementation for each actor Props.
the code like this:
```
//get current header ,otherwise return emptyMessageHeader
func (ctx *localContext) currentHeaderOrDefault() messageHeader {
	envelope, ok := ctx.message.(*MessageEnvelope)
	if ok {
		return envelope.Header
	}
	return emptyMessageHeader
}

func (ctx *localContext) Request(pid *PID, message interface{}) {
	env := &MessageEnvelope{
		Header:  ctx.currentHeaderOrDefault(), //<----change this
		Message: message,
		Sender:  ctx.Self(),
	}
	ctx.sendUserMessage(pid, env)
}

func (ctx *localContext) RequestFuture(pid *PID, message interface{}, timeout time.Duration) *Future {
	future := NewFuture(timeout)
	env := &MessageEnvelope{
		Header:  ctx.currentHeaderOrDefault(),  //<----change this
		Message: message,
		Sender:  future.PID(),
	}
	ctx.sendUserMessage(pid, env)

	return future
}

func (ctx *localContext) sendUserMessage(pid *PID, message interface{}) {
	if ctx.outboundMiddleware != nil {
		if env, ok := message.(*MessageEnvelope); ok {
			ctx.outboundMiddleware(ctx, pid, env)
		} else {
			ctx.outboundMiddleware(ctx, pid, &MessageEnvelope{
				Header:  ctx.currentHeaderOrDefault(),  //<----change this
				Message: message,
				Sender:  nil,
			})
		}
	} else {
		pid.ref().SendUserMessage(pid, message)
	}
}
```
Hope everyone can discuss this issue！

PotterDai:
Fixed here https://github.com/AsynkronIT/protoactor-go/pull/206

Issue #196, opened 2017-11-11
emptyMessageHeader should not be a Global Var


feyman2016:
Hi,I'm working on tracing with zipkins and I found that emptyMessageHeader  is currently a global var, and it is used by `context.Tell` , `pid.Tell` , `Request` and `RequestFuture` methods, when I modify and propagate the global variable `MessageHeader` between actors, you can imagine it would mess up.

For now, If I change the 
```
func (ctx *localContext) Tell(pid *PID, message interface{}) {
	ctx.sendUserMessage(pid, message)
}

func (ctx *localContext) sendUserMessage(pid *PID, message interface{}) {
	if ctx.outboundMiddleware != nil {
		if env, ok := message.(*MessageEnvelope); ok {
			ctx.outboundMiddleware(ctx, pid, env)
		} else {
			ctx.outboundMiddleware(ctx, pid, &MessageEnvelope{
				Header:  emptyMessageHeader,
				Message: message,
				Sender:  nil,
			})
		}
	} else {
		pid.ref().SendUserMessage(pid, message)
	}
}
```

to

```
func (ctx *localContext) Tell(pid *PID, message interface{}) {
	ctx.sendUserMessage(pid, message)
}

func (ctx *localContext) sendUserMessage(pid *PID, message interface{}) {
	if ctx.outboundMiddleware != nil {
		if env, ok := message.(*MessageEnvelope); ok {
			ctx.outboundMiddleware(ctx, pid, env)
		} else {
			ctx.outboundMiddleware(ctx, pid, &MessageEnvelope{
				Header:  make(messageHeader),
				Message: message,
				Sender:  nil,
			})
		}
	} else {
		pid.ref().SendUserMessage(pid, message)
	}
}
```
it worked. I didn't test `Request` and `RequestFuture`, but I think the situation should be the same.
What do you think of this?@rogeralsing

alex023:
At present, the method provided by the context only Tell, Request is not enough to meet the environment variable delivery. 
See also #200

alex023:
Sometimes, the default global variable is necessary, but the variable name `emptyMessageHeader`will be troubling.
In addition, the method provided by context only `Tell`,` Request` is not enough to meet the transfer of private environment variables, see #200

bryanpkc:
[...]

Issue #200, opened 2017-11-22
method "Forward" should be provided in actor.Context


alex023:
## Question
 In the actual business, we face the message there will be several situations:
1. Do not need reply:  
  pipe: actor.A--->actor.B---->actor.C
  We can use `ctx.Tell`
2. Do need reply:      actor.A---->actor.B--->actor.A
We can use `ctx.Request`
3. Do need reply by last actor:
actor.A---->actor.B---->actor.C---->actor.A
if we want transfer env by envelop.header ，what should actor.B adapt method?

## Suggest
I think ctx.Forward is necessary
the code may be:
```go
//context.go
type Context interface{
       //...
       //Forward sends a message to the given PID and also provides message orginal Sender PID
	Forward(pid *PID ,message interface{})
       //...
}
```
```go
//local_context.go
//Forward sends a message to the given PID and also provides message orginal Sender PID
func (ctx *localContext) Forward(pid *PID, message interface{}) {
	env := &MessageEnvelope{
		Header:  ctx.messageHeader(),
		Message: message,
		Sender:  ctx.Sender(),
	}
	ctx.sendUserMessage(pid, env)
}

func (ctx *localContext) messageHeader() messageHeader {
	envelope, ok := ctx.message.(*MessageEnvelope)
	if ok {
		return envelope.Header
	}
	return emptyMessageHeader
}
```
Looking forward to a better way！

rogeralsing:
Closed by #203

PR #203, opened 2017-11-23
implements for #201 #202

alex023:
implements for #201 #202, can meet #196
1. Context can pass current.Header if it is not nil, other wise pass default emptyHeader.
2. Pass the original Sender of the message through the new Forward method

### DIFF
--- a/router/common_test.go
+++ b/router/common_test.go
@@ -85,6 +85,10 @@ func (m *mockContext) Tell(pid *actor.PID, message interface{}) {
 	m.Called()
 }
 
+func (m *mockContext) Forward(pid *actor.PID, message interface{}) {
+	m.Called()
+}
+
 func (m *mockContext) Request(pid *actor.PID, message interface{}) {
 	m.Called()
 }


actor/local_context.go
```diff
@@ -85,7 +101,12 @@ func (ctx *localContext) MessageHeader() ReadonlyMessageHeader {
 }
 
 func (ctx *localContext) Tell(pid *PID, message interface{}) {
-	ctx.sendUserMessage(pid, message)
+	env := &MessageEnvelope{
```
rogeralsing: This will end up allocating a message envelope for each Tell, which will have a big impact on high throughput systems. 
The sendUserMessage func already do this envelope wrapping if needed, was there a specific reason why it was added directly to the Tell func also?

alex023: just as #200 mentioned , if in the same node, we want to transfer msg with private header through `actor.A`, `actor.B`... `actor.Z`, and finally, `actor.Z` respond to `actor.A`.
We can not pass private Header information just use `Context.Tell` or `Context.Request`. The heads may be out of business for `Zipkin` or other information, and have little to do with specific business.

rogeralsing: I understand the need for `Forward`, but I don't understand why the envelope needs to be created in the Tell or Request methods.
They both call into sendMessage which already does this.

PotterDai:
Implemented in https://github.com/AsynkronIT/protoactor-go/pull/224 and https://github.com/AsynkronIT/protoactor-go/pull/206

PR #205, opened 2017-11-27
Check message type is error or not in Future.

PotterDai:
@rogeralsing  This was brought up in Gitter chat. I'm not sure if this is the right way to do. What do you think?

### DIFF
--- a/actor/future.go
+++ b/actor/future.go
@@ -113,12 +113,12 @@ type futureProcess struct {
 
 func (ref *futureProcess) SendUserMessage(pid *PID, message interface{}) {
 	msg, _ := UnwrapEnvelope(message)
-	ref.result = msg
+	ref.processMessage(msg)
 	ref.Stop(pid)
 }
 
 func (ref *futureProcess) SendSystemMessage(pid *PID, message interface{}) {
-	ref.result = message
+	ref.processMessage(message)
 	ref.Stop(pid)
 }
 
@@ -139,6 +139,14 @@ func (ref *futureProcess) Stop(pid *PID) {
 	ref.runCompletions()
 	ref.cond.L.Unlock()
 	ref.cond.Signal()
+}
+
+func (ref *futureProcess) processMessage(message interface{}) {
+	if errMsg, ok := message.(error); ok {
+		ref.err = errMsg
+	} else {
+		ref.result = message
+	}
 }
 
 //TODO: we could replace "pipes" with this


rogeralsing:
I'm not sure tbh, this implies that someone sends an error back in a ctx.Respond, right? there are no other cases where this could happen?

Incase someone needs to reply with a negative response, I think that should be modelled as an explicit message.

But I'm not sure I have the complete background on this

PotterDai:
@rogeralsing  Agreed on an explicit message. From my understanding from @gviedma 's gitter chat Future should be able to handle error message other than timeout, so that if a request is errored, it can return that error message to the Future instead of let it timeout. That's the main purpose of implementing error in Future.

rogeralsing:
Ok, important to take into account is that futures might get responses from remote nodes, thus the remote actor will not be able to send an actual error object (needs to be serializable).
Also, we share contracts with C#, Kotlin, which makes the built in support for errors less useful IMO.

If anything, I'd rather see that we have a generic Failure message with e.g. string reason or something like that. then have a helper function to translate errors into that message.

Thoughts?

PotterDai:
@rogeralsing Completely agree! Also considering when we send context.Respond, in no way the sender will know the receiver is a Future or not, a generic failure message should be used and be adoped in the entire actor system.
Shall we just implement a ActorError proto message like this:

message ActorError {
    int32 error_code = 1;
    string error_message = 2;
}

And if we do implement this message type, should it be treated as SystemMessage?

bryanpkc:
@PotterDai Is your proposed `ActorError` only intended to capture Protoactor errors, e.g. unhandled issues in the remote/cluster packages? If so, it should be named a bit more specifically, e.g. `RuntimeError` or `ProtoactorError`. Right now it seems a bit too much like enforcing a protocol on what should be user-specified behavior.

rogeralsing:
I think it should just be a user message as it will be used like a normal response to other actors.
Also, ActorError might imply there was an error with the actor, but I suppose the same message might be used just as a negative response. eg.

```
if someCond {
   ctx.respond(something)
} else {
  ctx.respond(&Failure{ error_message = "some reason" }
}
```

So some more neutral name would be good I think

PotterDai:
@bryanpkc  @rogeralsing  Yep! Agree
@rogeralsing It seems Failure message is already used.
Shall we use "ProtoactorError" bryanpkc proposed? Or "ErrorResponse"/"FailureResponse"?

PotterDai:
@rogeralsing I will close this PR for now. After further implementing, I think this logic will very likely misguide user. Actor model should promote logic of "let it crash" instead of heavy error handling, thus we should not provide a common message to let user ctx.respond(err). 

In some cases where errors indeed need to be returned, the ideal way is to just follow what we did in cluster's ActorPidResponse, let user create a custom response message to return the custom error.

PR #208, opened 2017-12-11
Revise GrainCallOptions to follow dotnet version

PotterDai:
@rogeralsing 
This pr is to follow DotNet version's GrainCallOptions.

The good side:
-GrainCallOptions can be reused instead of allocating new one on every grain call.
-Cleaner API. Currently, user have to use cluster.WithTimeout/cluster.WithRetry, which might be confusing since we also have cluster config now. In this PR user can create GrainCallOptions by calling cluster.NewGrainCallOptions().WithTimeout().WithRetry()

The bad side:
-Golang doesn't have optional parameter support, so we will have 2 more apis for each grain method.
    client.SayHello/client.SayHelloWithOpts/client.SayHelloChan/client.SayHelloChanWithOpts.

What do you think about the change?


rogeralsing:
I think the extra api is fine as the client is code generated anyway

PotterDai:
@rogeralsing  Great! Then this pr should be good to go = )

PR #215, opened 2018-01-08
FromTemplate: produce an actor that resets correctly

### LINKED ISSUE #212: the actor instance is still reuse even after panic restarting
when actor panic, the instance is reused? This is different from akka .

- protoActor:clean chidren actor, except variable of instance
- AKKA: all reset

I found reason from *props.go*:
```go
func makeProducerFromInstance(a Actor) Producer {
	return func() Actor {
		return a
	}
}
```
is it problem?

### go for protoactor:
```go
package main

import (
	"fmt"
	"time"
	"github.com/AsynkronIT/protoactor-go/actor"
	"reflect"
)

type parentActor struct {
	count int  //if panic it will not reset
}

func (pa *parentActor) Receive(context actor.Context) {
	pa.count++
	msgType := reflect.TypeOf(context.Message())
	fmt.Printf("%v: receive message[%v] \n", pa.count,msgType)

	switch msg := context.Message().(type) {
	case *actor.Started:
		fmt.Println("   Started, initialize actor here")
	case *actor.Stopping:
		fmt.Println("   Stopping, actor is about shut down")
	case *actor.Stopped:
		fmt.Println("   Stopped, actor and it's children are stopped")
	case *actor.Restarting:
		fmt.Println("   Restarting, actor is about restart")
	case MessageInit:
		fmt.Println("   receive MessageInit")
		context.Spawn(actor.FromInstance(&childActor{}))
	case MessageCaculate:
		for _, child := range context.Children() {
			child.Request(msg, context.Sender())
		}
	case MessagePanic:
		fmt.Println("   receive MessagePanic")
		panic("receive MessagePanic")
	}
}

type childActor struct{}

func (ca *childActor) Receive(context actor.Context) {
	switch msg := context.Message().(type) {
	case MessageCaculate:
		fmt.Printf("   childActor caculate:%v multi %v =%v \n", msg.x, msg.y, msg.x*msg.y)
	}
}

type MessageInit struct{}
type MessagePanic struct{}
type MessageCaculate struct {
	x int
	y int
}

func main() {
	props := actor.FromInstance(&parentActor{})
	actor := actor.Spawn(props)

	actor.Tell(MessageInit{})
	time.Sleep(1 * time.Second)

	actor.Tell(MessageCaculate{2, 7})
	time.Sleep(1 * time.Second)

	actor.Tell(MessagePanic{})
	time.Sleep(1 * time.Second)

	actor.Tell(MessageCaculate{8, 2})
	time.Sleep(1 * time.Second)

	fmt.Println("read sent send stop")
	actor.Stop()

	//console.ReadLine()
}
```

### scala for akka
```scala
import java.util.concurrent.TimeUnit
import akka.actor.{Actor, ActorSystem, Props}

case class MessageCaculate(x: Int, y: Int)

case object MessageInit

case object MessagePanic

class ChildChild extends Actor {
  override def receive: Receive = {
    case MessageCaculate(x, y) =>
      println(s"   childActor caculate:$x multi $y =${x * y}")

  }
}

class ParentActor extends Actor {
  private var counter: Int = _ //it will reset if actor panic

  def getCounter: Int = {
    counter += 1
    counter
  }

  override def receive: Receive = {
    case MessageInit =>
      context.actorOf(Props(new ChildChild()), "child")
    case msg: MessageCaculate =>
      context.child("child") match {
        case Some(e) => e ! msg
        case None =>
          println("haven't child")
      }
    case MessagePanic =>
      7 / 0
  }

  override def postStop(): Unit = {
    println(getCounter + "  actorTester stoped!")
  }

  override def preStart(): Unit = {
    println(getCounter + "  actorTester preStart!")
  }

  override def postRestart(reason: Throwable): Unit = {
    println(getCounter + "  actorTester postRestart!")
  }
}

object Test extends App {
  val system = ActorSystem("tester")
  val tester = system.actorOf(Props[ParentActor])

  tester ! MessageInit
  TimeUnit.SECONDS.sleep(1)

  tester ! MessageCaculate(2, 7)
  TimeUnit.SECONDS.sleep(1)

  tester ! MessagePanic
  TimeUnit.SECONDS.sleep(1)

  tester ! MessageCaculate(8, 2)
  TimeUnit.SECONDS.sleep(2)

  system.terminate()
}
```

bryanpkc:
The new function FromTemplate can be used to create actors by copying from
an instance (known as the template for the actor). Each time the actor is
restarted, its state is reset to the contents of the template. This is
more intuitive than FromInstance, which should be reserved for immutable
value-type actors (and has been documented as such).

Fixes #212.

### DIFF
--- a/actor/props.go
+++ b/actor/props.go
@@ -1,6 +1,10 @@
 package actor
 
-import "github.com/AsynkronIT/protoactor-go/mailbox"
+import (
+	"reflect"
+
+	"github.com/AsynkronIT/protoactor-go/mailbox"
+)
 
 type InboundMiddleware func(next ActorFunc) ActorFunc
 type OutboundMiddleware func(next SenderFunc) SenderFunc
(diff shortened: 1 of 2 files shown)


bryanpkc:
@rogeralsing @PotterDai Any comment on this pull request?

PotterDai:
@bryanpkc Hey! Sorry for the late reply, was really busy with my own stuffs. = )
Really appreciate for the PR! However I'm still not so convinced about whether it is really necessary for us to have "FromInstance" and "FromTemplate" methods.
Could you elaborate some more use-cases for those 2 methods, other than saving a few lines of codes?
What do you think @rogeralsing ?

rogeralsing:
I'm leaning towards what @PotterDai says.
I'm not convinced the convenience outweighs the potential confusion.

bryanpkc:
The goal of this PR is to reduce confusion. I don't see how it will add potential confusion.

1. `FromInstance` was designed from day 1 to reuse the same instance for an actor as it reincarnates. This was never made clear in the godoc, which confused users.

2. `FromTemplate` makes a copy of the given object, which is what people usually expects when they try to use `FromInstance`. Adding this function and using it in our example programs would make it clear how an actor should be created from a template. I could push a commit to update the examples, if that helps.

3. Since many programs are already using `FromInstance` and may even be relying on its behavior, changing its semantics under the cover or removing it from the API could cause more confusion for users, and I advise against doing so.

PotterDai:
@bryanpkc Many programs are already using "FromInstance" doesn't mean that we cannot deprecate it if it causes confusion for future/current users. Adding documentation is more like a patch from my POV, it still creates an exception, making Actors created from "FromInstance" behaving differently with other methods like "FromFunc" / "FromProducer". This is not very intuitive and user friendly.

What I mean is that, if user do want to reuse an instance, user can do it like this.
instanceObj := &InstanceObj{...}
actor.FromProducer(func() actor.Actor {
    return &HoldingActor{instanceObj}
})
This way it differentiate instance object from its holding Actor, making things much clear.

PR #240, opened 2018-07-01
Fix problem with detection of endpoint termination

mhennings:
When a endpoint is terminated a EndpointTerminatedEvent is emitted using the eventstream.

When this message is processed it seems to be wrapped in a slice. This leads to a processing inside of sendEnvelopes which causes the actor to fail and be restarted. 

To avoid this problem I added a additional handling inside of sendEnvelopes. This can only be considered as a temporary solution. It seems that the event should not be wrapped in the first place.

### DIFF
--- a/remote/endpoint_writer.go
+++ b/remote/endpoint_writer.go
@@ -91,6 +91,13 @@ func (state *endpointWriter) sendEnvelopes(msg []interface{}, ctx actor.Context)
 	var targetID int32
 	var serializerID int32
 	for i, tmp := range msg {
+
+		switch unwrapped := tmp.(type) {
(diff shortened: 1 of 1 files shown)


mhennings:
To reproduce the problem start a cluster with random ports and send some messages. 
Terminate one member and restart it with a different port.

As a result the endpoint never stops to try to reconnect.

rogeralsing:
Good catch,
The EndpointWriter uses a batching mailbox, thus any message sent to it will be wrapped in a slice.
(except for ProtoActor built in messages that are not delivered via the mailbox)

I will look into this deeper to see what can be done about the whole process of remoting

PR #243, opened 2018-07-12
Add an SSL demo for the examples

shaneutt:
This adds a simple example of using SSL for inter-node communications.


PotterDai:
@shaneutt hey really sorry didn't notice this PR. Thank you very much and merge right the way!

PR #244, opened 2018-07-16
extract message from envelope to determine NotInfluenceReceiveTimeout

anopheles:


rogeralsing:
Thank you for spotting this!

PR #247, opened 2018-08-02
Corrected error in the lifecycle example

avneesh91:
There was a error in the example for illustrating the lifecycle of a actor.


avneesh91:
Issue: #248

PR #251, opened 2018-08-08
Follow latest major changes in ProtoActor-DotNet

PotterDai:
@rogeralsing 
So this PR is to follow latest major changes made in dotnet version.


rogeralsing:
Some compilation/test errors

PotterDai:
Yep, it's still in progress...

PotterDai:
Will start to work on this again in the following days, hopefully I can complete it before Christmas.

PotterDai:
@rogeralsing Hey Roger, the persistence / plugin module does not exist in DotNet version.
The issue is that plugin is designed as a ReceiverMiddleware, which will only be provided with actor.ReceiverContext instead of actor.Context after changes.

Now I just brutally cast actor.ReceiverContext to actor.Context (https://github.com/AsynkronIT/protoactor-go/blob/e5a9d9a264b01d5f190b2f9b5c3838efebe660d1/plugin/passivation.go#L54) so that  it can access ctx.Actor()/ctx.Self() values.

Not sure if it's the right way to do it.

PotterDai:
@rogeralsing Hey Roger, how do you think we should merge this change? Since it's a major update and will break all existing code base.

emilingerslev:
Would be great with all this merged soon 👍 Maybe as a new major version?

emilingerslev:
Really needs the new RootContext, since I keep getting into trouble with initiating requests on the edges of my application and getting the middleware to work at the same time. As far as I can see, these awesome improvements you've been making is resolving most of my issues.

Any news on a release of this? Maybe a release candidate or something?

actor/context.go
```diff
@@ -38,24 +46,39 @@ type Context interface {
...
-	// PushBehavior pushes the current behavior on the stack and sets the current Receive handler to the new behavior
-	PushBehavior(behavior ActorFunc)
+	AwaitFuture(f *Future, continuation func(res interface{}, err error))
+}
 
-	// PopBehavior reverts to the previous Receive handler
-	PopBehavior()
+type SenderContext interface {
+	// Message returns the current message to be processed
+	Message() interface{}
```
emilingerslev: Shouldn't the `Message()` be a part of the `ReceiverContext`? It seems strange to have a `Message()`on  RootContext. The same goes for the `MessageHeader()`

actor/root_context.go
```diff
@@ -0,0 +1,116 @@
...
+	senderMiddleware SenderFunc
+	headers          messageHeader
+}
+
+var emptyRootContext = &RootContext{
+	senderMiddleware: nil,
+	headers:          emptyMessageHeader,
+}
+
+func EmptyRootContext() *RootContext {
```
emilingerslev: Would be great if the `RootContext` wasn't something too special, like right now. I got several places where I want either a `Context` or a `RootContext` because I want to be able to Send or Spawn.
I played with adding an `OuterContext` like this

```
type OuterContext interface {
	SpawnerContext
	SenderContext
}
```

which I use instead of `RootContext` and that solves most of my problems, since that allows for using either a normal Context or a RootContext, while still being able to both Spawn and Send.

actor/root_context.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	// Default path
+	pid.sendUserMessage(message)
+}
+
+//
+// Interface: SpawnerContext
+//
+
+func (rc *RootContext) Spawn(props *Props) (*PID, error) {
```
emilingerslev: The signature on this should be
```suggestion
func (rc *RootContext) Spawn(props *Props) *PID {
```

actor/root_context.go
```diff
@@ -0,0 +1,116 @@
...
+	name := ProcessRegistry.NextId()
+	return rc.SpawnNamed(props, name)
+}
+
+func (rc *RootContext) SpawnPrefix(props *Props, prefix string) (*PID, error) {
```
emilingerslev: The signature on this should be

```suggestion
func (rc *RootContext) SpawnPrefix(props *Props, prefix string) *PID {
```

examples/cluster-metrics/member/main.go
```diff
@@ -19,25 +19,25 @@ const (
...
 // Logger is message middleware which logs messages before continuing to the next middleware
-func Logger(next actor.ActorFunc) actor.ActorFunc {
-	fn := func(context actor.Context) {
-		switch context.Message().(type) {
+func Logger(next actor.ReceiverFunc) actor.ReceiverFunc {
+	fn := func(context actor.ReceiverContext, env *actor.MessageEnvelope) {
+		switch env.Message.(type) {
 		case *actor.Started:
-			log.Printf("actor started " + context.Self().String())
+			log.Printf("actor started " + context.(actor.Context).Self().String())
```
emilingerslev: I got this exact same problem (casting) with the `ReceiverContext` in some of my middleware. Maybe it would make sense to extend the `ReceiverContext` with the some of the properties like `Self`?
I could also be the `SenderMiddleware` should use `Context` instead, since that would allow for more powerful Middleware solutions?

actor/root_context.go
emilingerslev: Could be that this `OuterContext` could be called `InitiatingContext`, since it's always about getting stuff started

actor/root_context.go
```diff
@@ -0,0 +1,116 @@
...
+	rc.sendUserMessage(pid, message)
+}
+
+func (rc *RootContext) Request(pid *PID, message interface{}) {
+	rc.sendUserMessage(pid, message)
+}
+
+func (rc *RootContext) RequestWithCustomSender(pid *PID, message interface{}, sender *PID) {
+	env := &MessageEnvelope{
+		Header:  nil,
```
emilingerslev: Shouldn't this be this?:

```suggestion
		Header:  rc.headers,
```

actor/root_context.go
```diff
@@ -0,0 +1,116 @@
...
+		Sender:  sender,
+	}
+	rc.sendUserMessage(pid, env)
+}
+
+// RequestFuture sends a message to a given PID and returns a Future
+func (rc *RootContext) RequestFuture(pid *PID, message interface{}, timeout time.Duration) *Future {
+	future := NewFuture(timeout)
+	env := &MessageEnvelope{
+		Header:  nil,
```
emilingerslev: Shouldn't this be this?:

```suggestion
		Header:  rc.headers,
```

actor/root_context.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func (rc *RootContext) sendUserMessage(pid *PID, message interface{}) {
+	if rc.senderMiddleware != nil {
+		if envelope, ok := message.(*MessageEnvelope); ok {
+			// Request based middleware
+			rc.senderMiddleware(rc, pid, envelope)
+		} else {
+			// tell based middleware
+			rc.senderMiddleware(rc, pid, &MessageEnvelope{nil, message, nil})
```
emilingerslev: Shouldn't this be this?:

```suggestion
			rc.senderMiddleware(rc, pid, &MessageEnvelope{rc.headers, message, nil})
```

PotterDai:
@emilingerslev Hi, sorry for the delay, but it seems @rogeralsing is busy with something and we havn't fully discussed yet about how to merge this new change into the existing repo.
Again thank you very much for the feedbacks! I'll look into it and making the changes!

actor/root_context.go
PotterDai: @emilingerslev Hi, thank you for pointing out! I guess the problem here is due to a lack of detailed documentation, and I do agree that it is indeed a bit confusing and not user friendly at this part.

The `nil` here is just to follow the common context pattern, where `context.headers` is  reserved for the incoming message, while for assigning headers in outgoing message, it can only be done in sender middleware, not through context or actor.

This is also the reason why `headers` and `message` field are in `SenderContext` instead of `ReceiverContext`. Since in some sender middleware,  user need to access received message's header like tracing id, etc. 

Hope I explained it clearly! = )

examples/cluster-metrics/member/main.go
PotterDai: @emilingerslev Yep! I also mentioned this comment https://github.com/AsynkronIT/protoactor-go/pull/251#issuecomment-449430686 , this is a very annoying situation. I'm also trying to come up some better ways to handle this situation, but it seems Go doesn't allow duplicate interface property, so it's a bit tricky where to put it = P

actor/root_context.go
emilingerslev: Alright, makes sense is just my understanding that's a bit... off 😄 So to clarify; what are the RootContext headers for then?

examples/cluster-metrics/member/main.go
emilingerslev: @PotterDai so I tried to look into this a bit, intrigued by the problem 😊 I see that you mirrored the interfaces for the middleware in the dotnet implementation, so it seems like it intended by @rogeralsing. However I still would like to iterate a bit on this, since it seems like there is something here that could be improved.

About the limitation of go a way to go about it is splitting the context parts into separate interfaces and composition them only in one layer like this 

```go
package actor

import "time"

type Context interface {
	InfoContextPart
	ContextPart
	ReceiverContextPart
	SenderContextPart
}

type SenderContext interface {
	InfoContextPart
	SenderContextPart
}

type ReceiverContext interface {
	InfoContextPart
	ReceiverContextPart
}

type InfoContextPart interface {
	// Parent returns the PID for the current actors parent
	Parent() *PID

	// Self returns the PID for the current actor
	Self() *PID

	// Sender returns the PID of actor that sent currently processed message
	Sender() *PID

	// Actor returns the actor associated with this context
	Actor() Actor
}

// Context contains contextual information for actors
type ContextPart interface {
	// ReceiveTimeout returns the current timeout
	ReceiveTimeout() time.Duration

	// Returns a slice of the actors children
	Children() []*PID

	// Respond sends a response to the to the current `Sender`
	// If the Sender is nil, the actor will panic
	Respond(response interface{})

	// Stash stashes the current message on a stack for reprocessing when the actor restarts
	Stash()

	// Watch registers the actor as a monitor for the specified PID
	Watch(pid *PID)

	// Unwatch unregisters the actor as a monitor for the specified PID
	Unwatch(pid *PID)

	// SetReceiveTimeout sets the inactivity timeout, after which a ReceiveTimeout message will be sent to the actor.
	// A duration of less than 1ms will disable the inactivity timer.
	//
	// If a message is received before the duration d, the timer will be reset. If the message conforms to
	// the NotInfluenceReceiveTimeout interface, the timer will not be reset
	SetReceiveTimeout(d time.Duration)

	CancelReceiveTimeout()

	// Forward forwards current message to the given PID
	Forward(pid *PID)

	AwaitFuture(f *Future, continuation func(res interface{}, err error))
}

type SenderContextPart interface {
	// Message returns the current message to be processed
	Message() interface{}

	// MessageHeader returns the meta information for the currently processed message
	MessageHeader() ReadonlyMessageHeader

	// Send sends a message to the given PID
	Send(pid *PID, message interface{})

	// Request sends a message to the given PID and also provides a Sender PID
	Request(pid *PID, message interface{})

	// Request sends a message to the given PID and also provides a Sender PID
	RequestWithCustomSender(pid *PID, message interface{}, sender *PID)

	// RequestFuture sends a message to a given PID and returns a Future
	RequestFuture(pid *PID, message interface{}, timeout time.Duration) *Future
}

type ReceiverContextPart interface {
	Receive(envelope *MessageEnvelope)
}

type SpawnerContext interface {
	// Spawn starts a new child actor based on props and named with a unique id
	Spawn(props *Props) *PID

	// SpawnPrefix starts a new child actor based on props and named using a prefix followed by a unique id
	SpawnPrefix(props *Props, prefix string) *PID

	// SpawnNamed starts a new child actor based on props and named using the specified name
	//
	// ErrNameExists will be returned if id already exists
	SpawnNamed(props *Props, id string) (*PID, error)
}
```

That would, as far as I can see, work but it's still hard to find the right way to "cut" these contexts.

PotterDai: This makes sense! I'll make a change on this one, and on dotnet version as well! Thank you for the proposal! = D

PotterDai: @emilingerslev Just implemented in this commit:
https://github.com/AsynkronIT/protoactor-go/pull/251/commits/3ec042d2b3fc5849fe8243397b268780ae58749b

emilingerslev: Looks good @PotterDai 👍 this should solve quite a few "casts" in the middleware level 😃 And glad you liked the proposal. Was it possible to mimic the same behavior/setup in dotnet?

PotterDai: Just created a PR at dotnet version as well, it's much simpler, since it doesn't have the problem of duplicate properties = )

actor/root_context.go
Chumper: this! I wanted to contribute that exact piece but I would prefer to get this PR forward.

Chumper:
Can we also implement an `Unstash()` method? Restarting an actor get all stashed messages makes no sense for me. If I stash messages to a certain point, I would want to call unstash to get them instead of restarting the actor. What do you think?

Chumper:
Also I think that we can implement the Stash and Timeout functionality similiar to the persistence mixins. That would make the actor really slim and we can compose additional functionality with mixins.
But maybe that is something for another PR.

actor/root_context.go
PotterDai: @emilingerslev  @Chumper  Hmm, sorry but I wasn't able to put myself into the shoes for this request. Could you guys describe a bit more about why the error should be removed? Thank you very much!  = )

actor/root_context.go
PotterDai: @emilingerslev Hmm, strange...Did I implement something wrong? I just double checked the RootContext and it indeed implmenets context for spawner and sender.

PotterDai:
@Chumper I see what you mean. It makes sense for adding an Unstash method. I'll make a new PR for this one.

actor/root_context.go
Chumper: We have two methods for spawning an actor, either in the `root_context` or in the `context` of an actor.

In the context of an actor we do not return an error but just the `PID`:
https://github.com/AsynkronIT/protoactor-go/blob/match-dotnet/actor/context.go#L102

So it makes sense to match the returns for also spawning in the root context.

My reason for removing: As that will spawn an actor with a prefixed name and a unique id that is taken from the process registry there should be no case where the name is already taken.

If that is the case I would opt for a panic inside the method because then something else is wrong in the system.

actor/root_context.go
Chumper: I would also opt to tighten the root context. The root context only needs to be able to send and spawn.
Therefore it should only have the `sender` and `spawner` interface.

I would also opt for moving the message and messageHeader into a separated message interface. That would remove the ability to access a message from the root context.

actor/context.go
Chumper: I would opt for a separated message interface if possible. As we have sender and receiver that need to access messages we need that in a different place.

actor/root_context.go
Chumper: Ok, there could be edge cases where the `SpawnPrefix` method could return the same name as an already existing actor:
```
context.SpawnNamed(props, "Foo$2")
context.SpawnPrefix(props, "Foo")
```
or similar shenanigans, the question is do we want to catch that? If so, then we should add the error to all spawn methods.

PotterDai: @Chumper Thank you very much for the explanation! I opted in for panic option since calling SpawnNamed with a name like "Foo$2" should always be avoided, I just added this in documentation. This has been fixed in latest commits.

PotterDai:
@Chumper @emilingerslev  Thank you for the review! I think I'll go ahead and merge it.
@rogeralsing It should be OK from my testing. Let's see how it goes! = )

emilingerslev:
Awesome! Will start using it Monday! 😁

PR #261, opened 2018-12-03
Fix nil pointer exception for not yet initialised memberList

dennis-tra:
The lock obviously needs to come after the nil check. 

Sorry for the premature initial PR.


dennis-tra:
Just a friendly comment to push this PR ✌️:)

PR #266, opened 2018-12-17
fix for removing in slice of  pointers

alexpantyukhin:
In [this atricle](https://github.com/golang/go/wiki/SliceTricks) is described the case when 'the type of the element is a pointer' have a potential memory leak problem. The suggested soultion was implemented. @PotterDai what do you think about it?


PotterDai:
@alexpantyukhin Wow, good catch! Didn't even notice that.Thank you very much! =D

alexpantyukhin:
@PotterDai Yes :) thank you too! I found this hint recetly, I think if the order is not important, it's possible to avoid `Copy` method also, just swap the last one and `i` element and 're-length' slice. But I see that `GetAllMembers()` which shows the members outside. Is this important to keep the order? If no, I can prepare the new one PR.

PR #273, opened 2019-02-16
Move Stop()/Poison() from PID to Context.

PotterDai:


potterdai:
Merge follow dotnet version.

PR #276, opened 2019-02-19
Fix backpressure example nil pointer exception

dennis-tra:


neowulf:
Yikes, looks like it got fixed as part of #287.

dennis-tra:
@neowulf You're totally right. I'll close this one.

PR #277, opened 2019-02-19
Reregister service if TTL refresh failed

dennis-tra:
Occasionally we saw the error message
```
[CLUSTER] [CONSUL] Failure refreshing service TTL
```

We noticed that the service/member was removed from consul and therefore its TTL could not be refreshed. This would result in an endless loop of TTL update attempts. This PR implements a reregistration of the service and member in case of a TTL update failure. It also updates the `PassTTL` method to `UpdateTTL` as the first one is deprecated.


potterdai:
@dennis-tra Hi, sorry for such late reply...
I just merge this PR with latest dev, and made some minor changes. Wondering if it is good to merge? Thank you very much! = )

dennis-tra:
Looks good to me 👍

PR #278, opened 2019-02-19
Spawn middleware

emilingerslev:
I've been working on introducing OpenTracing Jaeger tracing to protoactor-go and one of the things that I found made it a lot easier is introducing a concept around SpawnMiddleware.
The SpawnMiddleware is triggered when an Actor (actually a Context) spawn a new child.

The Spawn middleware can be useful in many scenarios including propagation of other middleware, tracing and I'm sure much more.

I've also implemented a MiddlewarePropagator that supports getting middleware spread out
to full actor trees, with nothing but adding it to the root actor.

### DIFF
--- a/actor/props.go
+++ b/actor/props.go
@@ -179,7 +179,7 @@ func (props *Props) WithFunc(f ActorFunc) *Props {
 func (props *Props) WithSpawnMiddleware(middleware ...SpawnMiddleware) *Props {
 	props.spawnMiddleware = append(props.spawnMiddleware, middleware...)
 
-	// Construct the sender middleware chain with the final sender at the end
+	// Construct the spawner middleware chain with the final spawner at the end
 	props.spawnMiddlewareChain = makeSpawnMiddlewareChain(props.spawnMiddleware, func(id string, props *Props, parentContext SpawnerContext) (pid *PID, e error) {
 		if props.spawner == nil {
 			return defaultSpawner(id, props, parentContext)
--- a/actor/props.go
+++ b/actor/props.go
@@ -2,22 +2,22 @@ package actor
 
 import (
 	"errors"
-
 	"github.com/AsynkronIT/protoactor-go/mailbox"
 )
 
 // Props types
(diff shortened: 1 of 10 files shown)


actor/props.go
```diff
@@ -174,6 +176,20 @@ func (props *Props) WithFunc(f ActorFunc) *Props {
 	return props
 }
 
+func (props *Props) WithSpawnMiddleware(middleware ...SpawnMiddleware) *Props {
+	props.spawnMiddleware = append(props.spawnMiddleware, middleware...)
+
+	// Construct the sender middleware chain with the final sender at the end
```
Chumper:
```suggestion
	// Construct the spawner middleware chain with the final sender at the end
```

emilingerslev: Copy paste mistake 😅 thanks

PotterDai:
@emilingerslev Thank you very much for the PR! I was busy with my work and will check this out later today! Thx! = )

PotterDai:
@emilingerslev I think adding this line to 
`  - go get github.com/magiconair/properties`
file: https://github.com/AsynkronIT/protoactor-go/blob/dev/.travis.yml
should fix the travis error!
Looks great! Is it possible to share your use case for SpawnMiddleware? I'm very curious = )

emilingerslev:
@PotterDai Fixed the problem mentioned above. I actually just wanted to use `stretchr/testify/assert` 😅
To see how it could be used look at my OpenTracing draft PR: https://github.com/AsynkronIT/protoactor-go/pull/279

I've setup a `TracingMiddleware` solution that would enable tracing for all actors: https://github.com/AsynkronIT/protoactor-go/pull/279/files#diff-7083d8e2eec99ae013b6fe5c2c15c489
I'll start working on a test soon, that show how it will be used.

emilingerslev:
Will continue work on this as a part of https://github.com/AsynkronIT/protoactor-go/pull/279, since that actually includes the usage of the SpawnMiddleware, thus always validates that it works as expected.

PR #279, opened 2019-02-21
Add OpenTracing middleware

emilingerslev:
This PR is adding my proposal for doing OpenTracing in a useful and only a bit opinionated manner.

The code doesn't setup or use jaeger, since it isn't a requirement, but it works great with jaeger. I've tested it with [jaeger-client-go](https://github.com/jaegertracing/jaeger-client-go).

Will add some tests in upcoming commits

This implementation use a few sync maps to store local defined opentracing
span's across several function scopes.
A nicer solution would be appreciated, but looking to the dotnet counterpart
the same behavior is implemented using thread static variables which isn't
an option in go.

### DIFF
--- a/actor/actor_context.go
+++ b/actor/actor_context.go
@@ -358,7 +358,14 @@ func (ctx *actorContext) SpawnNamed(props *Props, name string) (*PID, error) {
 		panic(errors.New("Props used to spawn child cannot have GuardianStrategy"))
 	}
 
-	pid, err := props.spawn(ctx.self.Id+"/"+name, ctx.self)
+	var pid *PID
+	var err error
+	if ctx.props.spawnMiddlewareChain != nil {
+		pid, err = ctx.props.spawnMiddlewareChain(ctx.self.Id+"/"+name, props, ctx)
+	} else {
+		pid, err = props.spawn(ctx.self.Id+"/"+name, ctx)
+	}
+
 	if err != nil {
 		return pid, err
 	}
--- a/actor/actor_context_test.go
+++ b/actor/actor_context_test.go
@@ -15,13 +15,13 @@ func TestActorContext_SpawnNamed(t *testing.T) {
 	defer removeMockProcess(pid)
 
 	props := &Props{
-		spawner: func(id string, _ *Props, _ *PID) (*PID, error) {
+		spawner: func(id string, _ *Props, _ SpawnerContext) (*PID, error) {
 			assert.Equal(t, "foo/bar", id)
 			return NewLocalPID(id), nil
 		},
 	}
 
-	parent := &actorContext{self: NewLocalPID("foo")}
+	parent := &actorContext{self: NewLocalPID("foo"), props: props}
 	child, err := parent.SpawnNamed(props, "bar")
 	assert.NoError(t, err)
 	assert.Equal(t, parent.Children()[0], child)
@@ -204,7 +204,7 @@ func benchmarkActorContext_SpawnWithMiddlewareN(n int, b *testing.B) {
 		props = props.WithSenderMiddleware(middlwareFn)
 	}
 
-	parent := &actorContext{self: NewLocalPID("foo")}
+	parent := &actorContext{self: NewLocalPID("foo"), props: props}
 	for i := 0; i < b.N; i++ {
 		parent.Spawn(props)
 	}
--- a/actor/context.go
+++ b/actor/context.go
@@ -9,16 +9,19 @@ type Context interface {
 	senderPart
 	receiverPart
 	spawnerPart
+	messagePart
 }
 
 type SenderContext interface {
 	infoPart
 	senderPart
+	messagePart
 }
 
 type ReceiverContext interface {
 	infoPart
 	receiverPart
+	messagePart
 }
 
 type SpawnerContext interface {
@@ -72,15 +75,17 @@ type basePart interface {
 	AwaitFuture(f *Future, continuation func(res interface{}, err error))
 }
 
-type senderPart interface {
-	// Sender returns the PID of actor that sent currently processed message
-	Sender() *PID
-
+type messagePart interface {
 	// Message returns the current message to be processed
 	Message() interface{}
 
 	// MessageHeader returns the meta information for the currently processed message
 	MessageHeader() ReadonlyMessageHeader
+}
+
+type senderPart interface {
+	// Sender returns the PID of actor that sent currently processed message
+	Sender() *PID
 
 	// Send sends a message to the given PID
 	Send(pid *PID, message interface{})
--- a/actor/context.go
+++ b/actor/context.go
@@ -21,6 +21,11 @@ type ReceiverContext interface {
 	receiverPart
 }
 
+type SpawnerContext interface {
+	infoPart
+	spawnerPart
+}
+
 type infoPart interface {
 	// Parent returns the PID for the current actors parent
 	Parent() *PID
--- a/actor/middleware/opentracing/activespan.go
+++ b/actor/middleware/opentracing/activespan.go
@@ -1,0 +1,35 @@
+package opentracing
+
+import (
+	"fmt"
+	"github.com/AsynkronIT/protoactor-go/actor"
+	"github.com/opentracing/opentracing-go"
+	"sync"
+)
+
+var activeSpan = sync.Map{}
+
+func getActiveSpan(pid *actor.PID) opentracing.Span {
+	value, ok := activeSpan.Load(pid)
+	if !ok {
+		return nil
+	}
+	return value.(opentracing.Span)
+}
+
+func clearActiveSpan(pid *actor.PID) {
+	activeSpan.Delete(pid)
+}
+
+func setActiveSpan(pid *actor.PID, span opentracing.Span) {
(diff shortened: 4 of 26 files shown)


PotterDai:
@emilingerslev Thank you for making the opentracing example! Now it makes much more sense! I think I'll just merge this one since it already contains the spawncontext PR.

emilingerslev:
@PotterDai Just closed the the https://github.com/AsynkronIT/protoactor-go/pull/278 PR and continue work here, like you proposed. I found a few problems while writing an OpenTracing/Jaeger example (https://github.com/AsynkronIT/protoactor-go/pull/279/commits/01776c7cc9a4067e3b3c07ada4dee07e26df54dd#diff-aeef9410f740ffc5a98008d77dfa65cf), but most are them are already fixed. Initially I used the protoactor Router to do the example, but I found that the router doesn't work too well with the new spawn middleware, so I need to get my hands dirty with that stuff.
Do you know how the router functionality works? And how middleware, in general, are handled in router?

PotterDai:
@emilingerslev Hi Emil! Thank you for the changes!
I didn't use Router in my projects, so my understanding of Router might not be very thorough.
Basically router is just an actor. The only difference is that Router got its own spawner and process. That's because Router actor got its own `RouterState` directly in `RouterProcess`.
In my memory I don't think Router has any middleware, that's because the props of Router is instantiated within the library, and cannot be customized by users. I'm not quite sure if it is done purposely. For example this: https://github.com/AsynkronIT/protoactor-go/blob/dev/router/consistent_hash_router.go#L82

emilingerslev:
@PotterDai The problem I've experienced is the OpenTracing not working with the router. I'm not totally sure where that chain is broken, but if I could get you to try to run the `examples/jaegertracing/main.go` example (by following the `examples/jaegertracing/README.md`) and see that using a normal actor tree everything works.

However, using a router breaks my implementation in unexpected ways. You can run the `examples/jaegertracing/router/main.go` example and see that all messages are not connected anymore.

I'll have to dive a bit into the whole thing to know more, but I'd hoped that you could tell me how the router differs from normal actor trees 😅

emilingerslev:
I think the problem is that routers are spawned using the `props.WithSpawnFunc` which "ignores" the new spawn middleware. It doesn't seem like WithSpawnFunc is thought to be used for anything else than testing and the router, so I'll just have to implement SpawnMiddleware for these scenarios.

And maybe make a note on WithSpawnFunc docs, that it will require special handling of spawn middleware

PotterDai:
@emilingerslev Hi Emil! Could you push the latest commit so that I can check out `examples/jaegertracing/router/main.go` and probably I can help! Thx! = )

emilingerslev:
@PotterDai that was... embarrassing 😊 I've rebased on the latest dev changes and fixed the stuff that shouldn't use the EmptyRootContext.

Also I just introduced a WithGuardian on the RootContext. This makes it easier to do the SpawnMiddleware, but it also could makes us able to deprecate the `props.WithGuardian`, that anyway only can be used with the RootContext.

PotterDai:
@emilingerslev Haha, absolutely not, actually I should be blamed for writing a confusing API. 
I'll take a look at the Router part later today. Hopefully I can come up with something! = )

PotterDai:
@emilingerslev Just checked the Router example. It seems Request message go through normally (passing through 3 layers), but actor.Started part is acting weird, am I understanding it correctly?

@rogeralsing It seems that this is going to be another major shift in our API, and I was wondering what's your opinion towards this? 
From my understanding, this direction is more about moving all middleware setup into RootContext, by using SpawnMiddleware, so that all the spawning process will inherit the middleware from RootContext using this way.

rogeralsing:
I've been very much out of the loop.
I like the idea that the middlewares are inherited from the parent context.

Does that mean, that inside an actor and I want to spawn a child actor that also inherits middleware, I would call SpawnMiddleware there too?

So there are SpawnMiddleware and SpawnMiddlewareNamed* counterparts for everything?

tobowers:
I'm just noting my excitement on allowing easy opentracing here. In fact, it's in our backlog to add open tracing to an app of ours :).

I think this whole thing would be easier to implement if there was a normal go-style "context" that went along with the messages... Similar to the headers, but with interface{] value instead of string and designed for local use instead of sending over the wire.

dennis-tra:
Hey guys,
do you think this PR will be merged any time soon? I'm looking forward to a couple of changes it will introduce besides opentracing :)

emilingerslev:
I've pushed this for quite some time, so I guess we should just get back to it 😄 
What do you think we need to get this merged @PotterDai ?

emilingerslev:
> I've been very much out of the loop.
> I like the idea that the middlewares are inherited from the parent context.
> 
> Does that mean, that inside an actor and I want to spawn a child actor that also inherits middleware, I would call SpawnMiddleware there too?
> 
> So there are SpawnMiddleware and SpawnMiddlewareNamed* counterparts for everything?

Actually @rogeralsing, as far as I remember, we just implemented all middleware around the named parts, since we let an outer part autogenerate the (P)ID first, and then pass it through the middleware.

potterdai:
@emilingerslev Hi, sorry I was busy with other stuffs recently. I will have a look at this PR this weekend! =D Hopefully I can get it merged!

emilingerslev:
@PotterDai no problem 😊 I thought, that I had some work on the PR missing, but simply can't find out what 😄

emilingerslev:
@PotterDai any way I can help you getting through this PR?

potterdai:
@emilingerslev Sorry I didn't keep my promise... should have checked out the repo last week! I'll do it today! = ) Will get you feedback when I got some progress!

potterdai:
@emilingerslev Merged!  Thank you for the PR! Sorry it took such long time.... = )

emilingerslev:
Awesome! Thanks for merging it 👍 Hope everybody else loves the improvements 😊

Strum355:
How can this be used with registering known actors?
```go
import (
	"github.com/AsynkronIT/protoactor-go/actor"
	"github.com/AsynkronIT/protoactor-go/actor/middleware/opentracing"
	"github.com/AsynkronIT/protoactor-go/remote"
	protoGen "github.com/org/repo/app/silo/generated/app/silo/protobufs"
)

	remote.Register("TrackerGrain", actor.PropsFromProducer(func() actor.Actor {
		return &protoGen.TrackerGrainActor{}
	}).WithSpawnMiddleware(opentracing.TracingMiddleware()))
```

This code compiles but the middleware doesnt seem to be invoked, the only mention of middleware along the callstack is at:

```go
protoactor-go/actor/actor_context.go

func (ctx *actorContext) processMessage(m interface{}) {
    if ctx.props.receiverMiddlewareChain != nil {
        ctx.props.receiverMiddlewareChain(ctx.ensureExtras().context, WrapEnvelope(m))
        return
    }

    if ctx.props.contextDecoratorChain != nil {
        ctx.ensureExtras().context.Receive(WrapEnvelope(m))
        return
    }

    ctx.messageOrEnvelope = m
    ctx.defaultReceive()
    ctx.messageOrEnvelope = nil // release message
}
```

emilingerslev:
@Strum355 I'm actually not too familiar with the `remote.Register` (I'm planning to get there), but I'd like to dive a bit into this. For one I think I should work on an example of using both spawn middlewares and the tracing.
For now, it would help a lot with a fully runnable example of above on a repo or in a gist. Can you help out with that?

rogeralsing:
@Strum355 it could be that the endpoint reader bypass some of the required infrastructure to pass the message to the target actor.
I'll have a look

Strum355:
@rogeralsing thatd be great if you could :) ive joined a project that uses protoactor-go and we're trying to introduce the tracing middleware to help us get insights into `future: timeout` spikes that we're having

rogeralsing:
I've been out of the loop for too long
@PotterDai SpawnMiddleware, does that apply to the direct actor or its children?

Because EndpointReader uses a non configured RootContext https://github.com/AsynkronIT/protoactor-go/blob/dev/remote/endpoint_reader.go#L79

rogeralsing:
So to answer my own question, I had a look at @emilingerslev's initial PR for this
https://github.com/AsynkronIT/protoactor-go/pull/278

Our remote spawned actors are spawned in a context lacking middleware.

Strum355:
@rogeralsing should or is it possible for them to get the middleware? Id assume that they should given that i can call the methods on them

emilingerslev:
I wasn't aware of this behavior when spawning actors on other nodes, but it makes sense. For this to work, I guess we need to add a way to setup the root context for the discussed EndpointWriter.
That way any actor spawned on that node will get fx OpenTracing.
And to answer your question @rogeralsing: the SpawningMiddleware was intended to add behavior to children. That way it also made it possible to make Spawning middleware that add itself to all children, and that way ensuring the middleware get destributed out to the whole tree.

rogeralsing:
What if different remote spawned actors need different middleware?

potentially we could extend the registration of remote props to allow to associate it to a specific context.

e.g.

remote.Register("someKind",props,rootContext); 
remote.Register("someOtherKind",props2,rootContext2); 

but then we need some way to fin that root context when an incoming message is read in the endpoint reader.

PR #280, opened 2019-02-21
Go modules support

emilingerslev:
This PR attempts to just add go modules support. This eliminates the need for installing dependencies in travis separately. Anybody not using modules, should not experience anything different since the new module files would just be ignored.

I can't see this should anything but improve the protoactor-go project.


emilingerslev:
@PotterDai What do you think about this? 😄

PotterDai:
@emilingerslev Great! Thx! I'll merge this right the way! =D

PR #281, opened 2019-02-26
Revise default values

PotterDai:
@rogeralsing 
This PR makes following changes:
1. make `EmptyMessageHeader` value public.
2. Change `EmptyRootContext` from a function to a value. Since a function confuses users that it will create a new RootContext every time it gets called, but in fact not.


PotterDai:
@rogeralsing It doesn't change any inside logic, so I'll merge it right the way. = )

Issue #282, opened 2019-02-28
can not print correct stack when actor panic


Qiaojiawei:
demo code:
```
package main

import (
	"fmt"
	"github.com/AsynkronIT/goconsole"
	"github.com/AsynkronIT/protoactor-go/actor"
)

func test() {
	var zero = 0
	_ = 1/zero
}


type hello struct{ Who string }
type helloActor struct{}

func (state *helloActor) Receive(context actor.Context) {
	switch msg := context.Message().(type) {
	case *hello:
		fmt.Printf("Hello %v\n", msg.Who)
		test()
	}
}

func main() {
	props := actor.PropsFromProducer(func() actor.Actor { return &helloActor{} })
	rootContext := actor.EmptyRootContext
	pid := rootContext.Spawn(props)
	rootContext.Send(pid, &hello{Who: "Roger"})
	console.ReadLine()
}
```

i divide zero in test function, when panic, it should print test function name, but the result as below:

```
Hello Roger
2019/02/28 07:09:18 [MAILBOX] [ACTOR] Recovering actor="nonhost/$1" reason="runtime error: integer divide by zero" stacktrace="main.(*helloActor).Receive:11" 
2019/02/28 07:09:18 [ACTOR] [SUPERVISION] actor="nonhost/$1" directive="RestartDirective" reason="runtime error: integer divide by zero"
```

as you can see, the top function is main.(*helloActor).Receive, not main.test,  when we have many files in project, its difficult to find where is problem.

PR #283, opened 2019-02-28
Fix: can not print correct stack function name when actor panic #282

Qiaojiawei:
when actor panic, it print top function name and line number, but  before there is a bug cause that function name is wrong when the name is little caption，as you can see #282 

after repaired, the function name is ok, i have past the result:
before:
old

after:
new


PotterDai:
@Qiaojiawei Good catch! Thank you very much! Merge right the way = )

Issue #264, opened 2018-12-05
Router - group: integer divide by zero


nikoszoisse:
Hello, I received a message `runtime error: integer divide by zero`
When I set up a router like this
```golang
			redisRouter, _ := context.
				SpawnNamed(
					router.NewRoundRobinGroup(),
					"redisRouter")
```
I solved it by not first creating the routees and then add them. 
e.g.
```golang
			var pids []*actor.PID
			for i := 0; i < int(msg.Count); i++ {
				props := actor.FromProducer(NewRedisActor)
				pid, _ := actor.SpawnPrefix(props, "rediscon_")
				pids = append(pids, pid)
			}
			redisRouter, _ := context.
				SpawnNamed(
					router.NewRoundRobinGroup(pids...).WithMiddleware(middleware.Logger),
					"redisRouter")
```
I will try to create a pull request and fix it if none does it until next week.

rogeralsing:
cc @PotterDai was this ever merged or is this still an issue?

nikoszoisse:
This issue it still exists. Although the way of doing such thing is wrong
and they very end.

Qiaojiawei:
the problem is roundRobinRoutee function use index%len(routees) to get an actor in the actor list, in this case len(routees)=0, it panic.
so we should check the length of rooutees and print error log?

PotterDai:
This should have been fixed in this PR:
https://github.com/AsynkronIT/protoactor-go/pull/284

PR #284, opened 2019-03-04
Fix: empty routees cause panic  #264

Qiaojiawei:
when we use
```
	rootContext := actor.EmptyRootContext
	pid, _ := rootContext.SpawnNamed(
			router.NewRoundRobinGroup(),
			"redisRouter")
	rootContext.Send(pid, &hello{Who: "Roger"})
```
or 
```
	pid, _ := rootContext.SpawnNamed(
			router.NewRandomGroup(),
			"redisRouter")
	rootContext.Send(pid, &hello{Who: "Roger"})
```
its cause panic, because router not check the length of routees, after repaired, it will print a log and ignore message.

default

### DIFF
--- a/router/random_router.go
+++ b/router/random_router.go
@@ -1,6 +1,7 @@
 package router
 
 import (
+	"log"
 	"math/rand"
 	"sync/atomic"
 	"unsafe"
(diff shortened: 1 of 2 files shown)


PotterDai:
@Qiaojiawei Hi sorry but it seems a quite legit fix! Just merged it, thank you very much! = )

PR #285, opened 2019-03-05
rootContext var is now only defined once per Service

otherview:
If you set up one or more Services in the protos.proto you'll get multiple declarations of the `rootContext` var.

Simple fix I guess :)


PotterDai:
Thank you for discovering this! Merge right the way = )

otherview:
Sweet! 👍

PR #287, opened 2019-03-07
fix(examples/backpressure): initialize the global context instead of local context

shiyanhui:


PotterDai:
@shiyanhui Thank you for finding it! = )

PR #290, opened 2019-03-13
set the supervisor strategy from the props

### LINKED ISSUE #289: Supervision nil pointer error
```
12:28 ~/code/go/src/github.com/AsynkronIT/protoactor-go (dev) $ go get ./...
13:56 ~/code/go/src/github.com/AsynkronIT/protoactor-go (dev) $ cd examples/supervision/
13:56 ~/code/go/src/github.com/AsynkronIT/protoactor-go/examples/supervision (dev) $ go run main.go 
Starting, initialize actor here
Hello Roger
2019/03/13 12:57:02 [MAILBOX] [ACTOR] Recovering actor="nonhost/$1/$2"Stopping, actor is about to shut down
 Stopped, actor and its children are stopped
reason="Ouch" stacktrace="main.(*childActor).Receive:40" 
2019/03/13 12:57:02 [MAILBOX] [ACTOR] Recovering actor="nonhost/$1" reason="runtime error: invalid memory address or nil pointer dereference" stacktrace="github.com/AsynkronIT/protoactor-go/actor.(*actorContext).handleFailure:501" 
2019/03/13 12:57:02 [ACTOR] [SUPERVISION] actor="nonhost/$1" directive="RestartDirective" reason="runtime error: invalid memory address or nil pointer dereference" 
```

I noticed this in our app when upgrading to the tip of dev, I went into the examples to see what was happening and found that it happens here too. I'm still looking into what's happening. It appears that the supervisor isn't actually getting attached to the context and triggering a nil here:

https://github.com/AsynkronIT/protoactor-go/blob/dev/actor/actor_context.go#L501

tobowers:
This seems to fix #289 . I couldn't see anywhere in the code where the supervisor is actually being set.

### DIFF
--- a/actor/actor_context.go
+++ b/actor/actor_context.go
@@ -100,8 +100,9 @@ type actorContext struct {
 
 func newActorContext(props *Props, parent *PID) *actorContext {
 	this := &actorContext{
-		parent: parent,
-		props:  props,
+		parent:     parent,
+		props:      props,
+		supervisor: props.supervisionStrategy,
 	}
 
 	this.incarnateActor()


PotterDai:
Thank you very much for this PR.
But the issue should be fixed here:
https://github.com/AsynkronIT/protoactor-go/pull/291
So I'll close this PR for now!

PR #294, opened 2019-03-16
Replaced Stop() with a system message

Chumper:
* Updated README.md with missing dependency
* Removed `Stop()` from the process interface and replaced it with a `stopMessage` system message 
* Fixed an error in the `proto-gen` package

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ go get github.com/Workiva/go-datastructures/queue
 go get github.com/emirpasic/gods/stacks/linkedliststack
 go get github.com/orcaman/concurrent-map
 go get github.com/AsynkronIT/gonet
+go get github.com/AsynkronIT/goring
 go get github.com/hashicorp/consul/api
 go get github.com/AsynkronIT/goconsole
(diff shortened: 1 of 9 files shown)


PotterDai:
Hi @Chumper 
Thank you for creating this PR! I agree that current state of Stop()/Poison() is indeed a bit confusing.
I created this PR before https://github.com/AsynkronIT/protoactor-go/pull/273 looking for moving those functions into a context as well. But I haven't discussed with other members yet.
Is this the same motive that you try to remove Stop() from pid as well?

Chumper:
My motive was to remove the stop message. The interface for the process does define the `Stop()`method but in my opinion it can be remmodeled as a system message. 
That's why I wanted to remove it.

I would also go one step further and thighten down the process interface with only a `Send` method, but that can be done later.

Is there any reason the `Stop()` method is defined here, that it can not be replacd with a system message? In my opinion that wasn't the case, so I removed it.

PotterDai:
@Chumper Aha! I see what you mean! Looks good! = )

PR #295, opened 2019-03-18
Fallback to default supervisor in handleFailure

brandonwestcott:
@PotterDai opening this as an extension to https://github.com/AsynkronIT/protoactor-go/pull/291. For cases when a supervisor has not been set explicitly, this makes the fallback `defaultSupervisionStrategy` rather than raising a nil pointer error.


potterdai:
@brandonwestcott Thank you very much for the fix! = )

PR #298, opened 2019-03-26
[cluster] Errors from the application are not re-tried now

otherview:
Hi guys!

I've tweak how the `{{ $method.Name }}WithOpts` is working.
Mainly because it would trigger the `opts.RetryAction` on **any** error.

If the error was a `future: timeout` or a legitimate application error, it would retry it `opts.RetryCount` times.

I don't think that was intended. The retry pattern for requesting data from the actor is solid, it retries if the actor is busy or some networking issue - but if the actor returns an legit error I don't see the point of retrying.

Main change was here:
```
for i := 0; i < opts.RetryCount; i++ {
		res, err = fun()
		if err == nil || err.Error() != "future: timeout" {
			return res, nil
		} else if opts.RetryAction != nil {
				opts.RetryAction(i)
		}
```

It's only retrying it the error is a `future: timeout` - I'm happy to extend to other cluster errors if you guys see fit. Or take some other approach that allows for application errors to go throught the cluster without being retried.

Thanks!

### DIFF
--- a/protobuf/protoc-gen-gograin/template.go
+++ b/protobuf/protoc-gen-gograin/template.go
@@ -54,6 +54,7 @@ func (g *{{ $service.Name }}Grain) {{ $method.Name }}(r *{{ $method.Input.Name }
 
 // {{ $method.Name }}WithOpts requests the execution on to the cluster
 func (g *{{ $service.Name }}Grain) {{ $method.Name }}WithOpts(r *{{ $method.Input.Name }}, opts *cluster.GrainCallOptions) (*{{ $method.Output.Name }}, error) {
+	result := &{{ $method.Output.Name }}{}
 	fun := func() (*{{ $method.Output.Name }}, error) {
 			pid, statusCode := cluster.Get(g.ID, "{{ $service.Name }}")
 			if statusCode != remote.ResponseStatusCodeOK {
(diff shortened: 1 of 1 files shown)


protobuf/protoc-gen-gograin/template.go
```diff
@@ -70,30 +71,29 @@ func (g *{{ $service.Name }}Grain) {{ $method.Name }}WithOpts(r *{{ $method.Inpu
...
+				return result, errors.New("unknown response")
 			}
 		}
 	
 	var res *{{ $method.Output.Name }}
 	var err error
 	for i := 0; i < opts.RetryCount; i++ {
 		res, err = fun()
-		if err == nil {
+		if err == nil || err.Error() != "future: timeout" {
```
potterdai, line 90: @otherview Thank you for the PR and really sorry for late reply! = )
I'm not so sure on this one. If there is an error, even if it's not the future: timeout one, we should still return nil instead of an empty object?

otherview: Good spot! Oversight when moving this from the test code.

That should be returning the err :) 👍

potterdai: @otherview Thank you for revision! = ) Just another question, is there any reason to move `result := &{{ $method.Output.Name }}{}` to the top, and return the empty object instead of `nil` when receiving error? Thx! = )

otherview: Good point.

I think my initial reasoning at first was to mimic the behaviour of protobuf where you always return something.
Now.. I don't think it adds much to be honest. 
Happy to revert if you think the same 👍

potterdai: @otherview Sure! Sorry I got so many requests, but I think `return nil` should suit the situation better here... Thank you very much!! = )

otherview: No problem @PotterDai and thanks for your input man :)

potterdai:
@otherview  Merged! = )

PR #301, opened 2019-04-02
Fix stop message handling for router actor #300

### LINKED ISSUE #300: Fix stop message handling for router actor
For example, parent has a child, which is a router actor. As we know it, router has two pids, one is router pid, another is proxy pid. Let's suppose parent pid is called `parent`, child actor has proxy pid `parent/child` and router pid `parent/child/router`.  And actually parent only knew child actor's proxy pid `parent/child`. 

When parent stops, it will stop the child actor. When router actor stops,  it will process following functions in router/process.go.

```
func (ref *process) Stop(pid *actor.PID) {
	if atomic.SwapInt32(&ref.stopping, 1) == 1 {
		return
	}

	ref.router.StopFuture().Wait()
	actor.ProcessRegistry.Remove(pid)
	ref.SendSystemMessage(pid, &actor.Stop{})
}
```

`ref.router.StopFuture().Wait()` will stop the router pid `parent/child/router`. And `parent/child/router` will send `parent` a Terminated message. Since `parent` only knew router's proxy pid `parent/child`, this Terminated message will do nothing. 

`ref.SendSystemMessage(pid, &actor.Stop{})` will send `parent/child` a Stop message. Then handle this Stop message as follows:
```
                case *actor.Stop:
		term := &actor.Terminated{Who: pid}
		ref.mu.Lock()
		ref.watchers.ForEach(func(_ int, other actor.PID) {
			if r, ok := actor.ProcessRegistry.Get(&other); ok {
				r.SendSystemMessage(&other, term)
			}
		})
		ref.mu.Unlock()
```
Since parent does not watch/unwatch child now, `parent` is not in `parent/child`'s watchers list anymore. Thus, `parent/child` won't send `parent` Terminated message. This will cause that  parent won't Stop. 

Here we propose this fix, to make sure that when parent stops, router actor will send parent Terminated message to stop normally.

yangdiangzb:
close #300

### DIFF
--- a/router/config.go
+++ b/router/config.go
@@ -93,5 +93,6 @@ func spawn(id string, config RouterConfig, props *actor.Props, parent *actor.PID
 		wg.Wait() // wait for routerActor to start
 	}
 
+	ref.parent = parent
 	return proxy, nil
 }
--- a/router/process.go
+++ b/router/process.go
@@ -10,6 +10,7 @@ import (
 // process serves as a proxy to the router implementation and forwards messages directly to the routee. This
 // optimization avoids serializing router messages through an actor
 type process struct {
+	parent   *actor.PID
 	router   *actor.PID
 	state    RouterState
 	mu       sync.Mutex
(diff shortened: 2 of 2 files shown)


potterdai:
Makes sense! Thank you very much for discover this!

PR #302, opened 2019-04-05
[cluster] Adding Grains configurable timeout lifecycle + terminate method

otherview:
Hi !

Added a couple of things I think grains should have - Based on using Orleans Grain lifecycle.

- Grain Life Time Out
When a grain is inactive for X amount of time it should be released from memory.
```
remote.Register("UserGrain", actor.PropsFromProducer(func() actor.Actor {
		timeout := 5 * time.Minute
		return &protoGen.UserGrainActor{
			Timeout: &timeout,
		}
	}))
```
After Timeout duration the UserGrain will poison itself.
```
case *actor.Started:
		a.inner = x{{ $service.Name }}Factory()
		id := ctx.Self().Id
		a.inner.Init(id[7:]) // skip "remote$"
		if a.Timeout != nil {
			ctx.SetReceiveTimeout(*a.Timeout)
		}
case *actor.ReceiveTimeout:
		ctx.Self().Poison()
```

- Terminate method to be run on Grain Life Time Out
Similar to `Init` a method that allows to run operations when the Grain is timing out.
```
// {{ $service.Name }} interfaces the services available to the {{ $service.Name }}
type {{ $service.Name }} interface {
	Init(id string)
	Terminate()
	{{ range $method := $service.Methods}}	
	{{ $method.Name }}(*{{ $method.Input.Name }}, cluster.GrainContext) (*{{ $method.Output.Name }}, error)
	{{ end }}	
}
```

and 
```
case *actor.ReceiveTimeout:
		a.inner.Terminate()
		ctx.Self().Poison()
```

Open to suggestions ! :)

### DIFF
--- a/protobuf/protoc-gen-gograin/template.go
+++ b/protobuf/protoc-gen-gograin/template.go
@@ -37,6 +37,7 @@ func Get{{ $service.Name }}Grain(id string) *{{ $service.Name }}Grain {
 // {{ $service.Name }} interfaces the services available to the {{ $service.Name }}
 type {{ $service.Name }} interface {
 	Init(id string)
+	Terminate()
 	{{ range $method := $service.Methods}}	
 	{{ $method.Name }}(*{{ $method.Input.Name }}, cluster.GrainContext) (*{{ $method.Output.Name }}, error)
 	{{ end }}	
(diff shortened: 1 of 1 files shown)


potterdai:
@otherview Sorry for late reply! Merge right the way! = )

PR #306, opened 2019-04-17
Print log messages in one piece

fabotronix:
This prevents interleaved output, if running in an environment with other writing go routines.


potterdai:
@fabotronix Thank you very much for the PR! Merge right the way = )

PR #311, opened 2019-05-07
[cluster] Adding PROCESSNAMEALREADYEXIST doesnt seem to be a failure

otherview:
Will need some review on this one. Not 100% if my interpretation is accurate.

In the cluster template, we have 
```
if statusCode != remote.ResponseStatusCodeOK {
```

Which seems to fail a lot with status code 3 -`ResponseStatusCodePROCESSNAMEALREADYEXIST`

From what I could tell this is a problem in the activator ?
Seems like it's trying to activate something that's already active.

Since the only time this error occurs the pid is returned
```
response := &ActorPidResponse{
				Pid:        pid,
				StatusCode: ResponseStatusCodePROCESSNAMEALREADYEXIST.ToInt32(),
			}
```

I've added to the template as an ignorable error. In my cluster haven't seem more error spikes around Error Code 3.

Sorry for the quick and dry PR, don't have much time today! Thanks!!


rogeralsing:
Good catch, yes the process already exists should not throw and is only an information that someone have concurrently spawned this process

PR #326, opened 2019-06-19
fix data race problems

### LINKED ISSUE #322: fix data race problems
Here are some data race problems in protoactor-go.
If we run `go test -race -v ./...`, we will find that data race are detected in many unit tests.
This issue will fix those data race problems.

yangdiangzb:
close #322

### DIFF
--- a/actor/actor_context_test.go
+++ b/actor/actor_context_test.go
@@ -2,6 +2,7 @@ package actor
 
 import (
 	"fmt"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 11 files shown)


stuartcarnie (approved):
Looks good to me

bryanpkc:
@yangdiangzb Could you please rebase your branch on dev to get rid of the spurious merge commit?

yangdiangzb:
> @yangdiangzb Could you please rebase your branch on dev to get rid of the spurious merge commit?

@bryanpkc Fixed.

potterdai:
Hi, sorry for such such late comments, but I am not sure about this PR after diving into the latest code base recently. I think some of the race problems mentioned in this PR lie within the testing methods rather than the functions themselves. 

Taking this line for example: https://github.com/AsynkronIT/protoactor-go/blob/69575e77b84f80cd785c01fb397634997f12129d/router/random_router.go#L25-L29

Essentially **randomRouterState** is the state of **RandomRouterActor**, which should be guaranteed to be thread safe. **SetRoutees** method will only be called here:

https://github.com/AsynkronIT/protoactor-go/blob/28110571d6a3931291bfc06476f23ec672337dd5/router/routeractor_pool.go#L23-L30

Adding the atomic pointer store to change the state of RouterActor is a bit redundant and it defies the purpose of using actor model, or am I missing something here? @rogeralsing @yangdiangzb @stuartcarnie

PR #328, opened 2019-06-21
Melaurent/add advertised address

gosthell:
I added an example too to demonstrate the change, but it might not be necessary


gosthell:
AsynkronIT/protoactor-go/issues/184

rogeralsing:
Very nice! thanks

PR #331, opened 2019-06-28
Resolve data race in internal/queue/mpsc

stuartcarnie:
Concurrent calls to Push and Pop result in a data race reported by the Go race detector


stuartcarnie:
This was fixed in #326

PR #333, opened 2019-07-08
Fix HandleFailure bug in all-for-one strategy

### LINKED ISSUE #332: Bug in all-for-one stragegy
https://github.com/AsynkronIT/protoactor-go/blob/3847eb264f0a2c4a275ecdf042f8875d346f2b5e/actor/strategy_all_for_one.go#L50
		
Is there any reason that we change `message` to `child`?

yangdiangzb:
close #332


bryanpkc:
LGTM.

PR #334, opened 2019-07-09
Fixed panic recovery loop on Unavailable GRPC error

Strum355:
When a member becomes unavailable (even before its picked up by Consul), trying to send messages to it (aka starting the endpoint writer) would cause a panic-recovery loop starting [here](https://github.com/AsynkronIT/protoactor-go/blob/dev/remote/endpoint_writer.go#L37) and recovered [here](https://github.com/AsynkronIT/protoactor-go/blob/dev/remote/endpoint_writer_mailbox.go#L80). After being escalated, the child being [restarted](https://github.com/AsynkronIT/protoactor-go/blob/d33cb30063583c13fceaf4ccd1418c32a47e0544/remote/endpoint_manager.go#L152) (which is an EndpointSupervisor) would again try to connect to the unavailable remote actor, rinse and repeat. Even if the unavailable node becomes available again, the loop still seems to occur from what I've seen.

Resuming the child rather than restarting it and allowing it to reinitialize gives us an eventual recovery (generally after 1-2 failed messages) and doesn't result in a panic-recovery loop.

The sequence of logs in the panic loop: `cluster.[REMOTE] - Started EndpointWriter`, `cluster.[REMOTE] - EndpointWriter connecting`, `cluster.[REMOTE] - EndpointWriter failed to connect`, `cluster.[REMOTE] - [ACTOR] Recovering`

Im sure there is a much better way to solve this issue that would also give us faster recovery, hopefully we can figure one out through discussion in this PR :) If more info is needed, logs etc, do let me know

### DIFF
--- a/cluster/consul/consul_provider.go
+++ b/cluster/consul/consul_provider.go
@@ -178,7 +178,7 @@ func (p *ConsulProvider) registerService() error {
 		Port:    p.port,
 		Check: &api.AgentServiceCheck{
 			DeregisterCriticalServiceAfter: p.deregisterCritical.String(),
-			TTL: p.ttl.String(),
+			TTL:                            p.ttl.String(),
 		},
 	}
 	return p.client.Agent().ServiceRegister(s)
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,9 @@ require (
 	github.com/AsynkronIT/goconsole v0.0.0-20160504192649-bfa12eebf716
 	github.com/AsynkronIT/gonet v0.0.0-20161127091928-0553637be225
 	github.com/Workiva/go-datastructures v1.0.50
+	github.com/chzyer/logex v1.1.10 // indirect
 	github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e
+	github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1 // indirect
 	github.com/codahale/hdrhistogram v0.0.0-20161010025455-3a0bb77429bd // indirect
 	github.com/couchbase/gocb v1.5.2 // indirect
 	github.com/emirpasic/gods v1.12.0
@@ -17,14 +19,17 @@ require (
 	github.com/hashicorp/go-cleanhttp v0.5.0 // indirect
 	github.com/hashicorp/go-rootcerts v1.0.0 // indirect
 	github.com/hashicorp/serf v0.8.2 // indirect
+	github.com/mitchellh/go-testing-interface v1.0.0 // indirect
 	github.com/mitchellh/mapstructure v1.1.2 // indirect
 	github.com/opentracing/opentracing-go v1.0.2
 	github.com/orcaman/concurrent-map v0.0.0-20190107190726-7ed82d9cb717
 	github.com/pkg/errors v0.8.1 // indirect
 	github.com/serialx/hashring v0.0.0-20180504054112-49a4782e9908
 	github.com/stretchr/testify v1.3.0
+	github.com/uber-go/atomic v1.4.0 // indirect
 	github.com/uber/jaeger-client-go v2.15.0+incompatible
 	github.com/uber/jaeger-lib v1.5.0
+	go.uber.org/atomic v1.4.0 // indirect
 	golang.org/x/net v0.0.0-20190213061140-3a22650c66bd
 	google.golang.org/grpc v1.18.0
 	gopkg.in/couchbase/gocbcore.v7 v7.1.11 // indirect
--- a/go.sum
+++ b/go.sum
@@ -10,8 +10,12 @@ github.com/armon/go-metrics v0.0.0-20180917152333-f0300d1749da h1:8GUt8eRujhVEGZ
 github.com/armon/go-metrics v0.0.0-20180917152333-f0300d1749da/go.mod h1:Q73ZrmVTwzkszR9V5SSuryQ31EELlFMUz1kKyl939pY=
 github.com/armon/go-radix v0.0.0-20180808171621-7fddfc383310/go.mod h1:ufUuZ+zHj4x4TnLV4JWEpy2hxWSpsRywHrMgIH9cCH8=
 github.com/bgentry/speakeasy v0.1.0/go.mod h1:+zsyZBPWlz7T6j88CTgSN5bM796AkVf0kBD4zp0CCIs=
+github.com/chzyer/logex v1.1.10 h1:Swpa1K6QvQznwJRcfTfQJmTE72DqScAa40E+fbHEXEE=
+github.com/chzyer/logex v1.1.10/go.mod h1:+Ywpsq7O8HXn0nuIou7OrIPyXbp3wmkHB+jjWRnGsAI=
 github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e h1:fY5BOSpyZCqRo5OhCuC+XN+r/bBCmeuuJtjz+bCNIf8=
 github.com/chzyer/readline v0.0.0-20180603132655-2972be24d48e/go.mod h1:nSuG5e5PlCu98SY8svDHJxuZscDgtXS6KTTbou5AhLI=
+github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1 h1:q763qf9huN11kDQavWsoZXJNW3xEE4JJyHa5Q25/sd8=
+github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1/go.mod h1:Q3SI9o4m/ZMnBNeIyt5eFwwo7qiLfzFZmjNmxjkiQlU=
 github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDkc90ppPyw=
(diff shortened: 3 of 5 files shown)


dennis-tra:
Hey @Strum355,

we encountered this error as well and I tried to fix it with PR #269. However, my fix lead to a DeadLock. I haven't had a thorough look at your changes, but I'm glad someone else is also having a look at this problem and trying to solve it. I'll definitely want to participate in the discussion and eventually get rid of this bug :)

Cheers,
Dennis

Strum355:
Hey @dennis-tra :)

We've had fairly good results with this solution over the past 24hrs, including a lil chaos engineering, bringing down one of two and two of two k8s pods to test this. The recovery period for every actor is fairly short depending on how many messages come through the cluster for it but youll probably end up seeing a lot of messages over time depending on how many actors you have until theyve all stabilised.

What i wonder is if one node goes down and the other node picks up the slack, does the cluster rebalance itself if/when the other node comes back online?

Theres definitely improvements that can be made to this fix, i discovered it by accident messing around with the different actions you can take on the child :sweat_smile:

dennis-tra:
Hey @Strum355,

This is our current solution to the problem:

We  gave the [remoteConfig](https://github.com/AsynkronIT/protoactor-go/blob/3d57163d4695b165890d69c4423f910af1c8e1d5/remote/config.go#L60) struct a new field called `connectionDecider` of type `func(string) bool`. It takes the address the endpoint writer wants to connect to and returns a boolean indicating whether it should connect to it or not:
```go
type remoteConfig struct {
	serverOptions            []grpc.ServerOption
	callOptions              []grpc.CallOption
	dialOptions              []grpc.DialOption
	endpointWriterBatchSize  int
	endpointWriterQueueSize  int
	endpointManagerBatchSize int
	endpointManagerQueueSize int
	connectionDecider        func(string) bool
}
```
With this approach it is possible to inject custom logic to the connection process.

In the [initialize](https://github.com/AsynkronIT/protoactor-go/blob/3d57163d4695b165890d69c4423f910af1c8e1d5/remote/endpoint_writer.go#L30) method of the `endpointWriter` we are calling the `connectionDecider` with the given address:
```go
func (state *endpointWriter) initialize() bool {

	if !state.config.connectionDecider(state.address) {
		plog.Info("EndpointWriter not connecting based on connection decider", log.String("address", state.address))
		terminated := &EndpointTerminatedEvent{
			Address: state.address,
		}
		eventstream.Publish(terminated)
		return false
	}

	err := state.initializeInternal()
	if err != nil {
		plog.Error("EndpointWriter failed to connect", log.String("address", state.address), log.Error(err))
		// Wait 2 seconds to restart and retry
		// Replace with Exponential Backoff
		time.Sleep(2 * time.Second)
		panic(err)
	}

	return true
}
```

Our `connectionDecider` implementation looks like this:
```go
remotingOption = append(remotingOption, remote.WithConnectionDecider(func(address string) bool {
	adrList := cluster.GetMemberAddresses()
	for _, v := range adrList {
		if v == address {
			return true
		}
	}
	log.Infof("Address %s is no known member. Deciding not to connect", address)
	return false
}))
```
This means when the address of the remote endpoint is not part of the cluster anymore (according to Consul) we won't connect to it. The other way around: If the remote endpoint is still available according to consul we will attempt to connect to it.

The new `GetMemberAddresses` method in the `cluster` package looks like this:
```Go
func GetMemberAddresses() []string {
	addresses := []string{}
	if memberList == nil {
		return addresses
	}

	memberList.mutex.RLock()
	defer memberList.mutex.RUnlock()

	for _, value := range memberList.members {
		addresses = append(addresses, value.Address())
	}
	return addresses
}
```

<details>
<summary>Some more details</summary>

We need to add some `nil`-checks because otherwise an uninitialized connection will be closed in the `*actor.Stopped` case - and we added the other checks "just in case"

```Go
func (state *endpointWriter) Receive(ctx actor.Context) {
	switch msg := ctx.Message().(type) {
	case *actor.Started:
		if !state.initialize() {
			ctx.Self().Stop()
		}
	case *actor.Stopped:
		if state.conn != nil {
			state.conn.Close()
		}
	case *actor.Restarting:
		if state.conn != nil {
			state.conn.Close()
		}
	case *EndpointTerminatedEvent:
		ctx.Stop(ctx.Self())
	case []interface{}:
		if state.stream != nil {
			state.sendEnvelopes(msg, ctx)
		}
	case actor.SystemMessage, actor.AutoReceiveMessage:
		// ignore
	default:
		plog.Error("EndpointWriter received unknown message", log.String("address", state.address), log.TypeOf("type", msg), log.Message(msg))
	}
}
```

The `WithConnectionDecider` method:
```Go
func WithConnectionDecider(decider func(string) bool) RemotingOption {
	return func(config *remoteConfig) {
		config.connectionDecider = decider
	}
}
```
</details>

Strum355:
Hey @dennis-tra , does this work even when the endpoint writer tries to connect in the time between when a node becomes unavailable and when Consul has been updated?

dennis-tra:
As long as consul has not been updated we are still running into the panic/restart loop. 

As soon as consul has cleared the node from its internal state we stop reconnecting, which would happen after 1 minute latest if I remember it right.

Strum355:
@dennis-tra do you have the link to the fork with these changes that we could potentially try out? @otherview

dennis-tra:
Here it is :)
https://github.com/dennis-tra/protoactor-go

rogeralsing:
Status here? are there additional fixes that needs to go in from @dennis-tra ?

Strum355:
We haven't gotten around to trying out those changes yet due to time constraints at work

dennis-tra:
@rogeralsing: Our approach is significantly different from the one proposed by @Strum355 and to discuss our implementation I opened the PR #344. I'll update the description soon.

---

Some comments to this PR:

1. As I see it the `endpointWriter` actor may run into a nil pointer exception when the member actually leaves the cluster and the actor is stopped. It is not guaranteed that the `state.conn` property of the `endpointWriter` in the `*actor.Stopped` and `*actor.Restarting` cases ([line](https://github.com/AsynkronIT/protoactor-go/blob/29182d2debd5aeb30fca0f3e6b0ae2c4ca26c839/remote/endpoint_writer.go#L162)) is properly initialised - this is something we also covered in our implementation and could be fixed with a simple nil check :)

2. Are you sure other errors should not be handled the same way?
  Just leaving this here: [I looked up](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) what the gRPC status code 14 (UNAVAILABLE) actually means:
   > The service is currently unavailable. This is most likely a transient condition, which can be corrected by retrying with a backoff. Note that it is not always safe to retry non-idempotent operations.

   From this I would conclude that it's perfectly valid to strive for an earlier reconnect instead waiting for the next message to be sent to the remote actor.

3. In you first post you mentioned:
   > Even if the unavailable node becomes available again, the loop still seems to occur from what I've seen.
   
   If a node was unavailable and became available again could it have received a new address and the `endpointWriter` still tried to connect to the old one?

otherview:
> @dennis-tra do you have the link to the fork with these changes that we could potentially try out? @otherview

Hi guys!

We're pretty happy with @Strum355 fix at the moment. We've confirmed that the error now corrects itself out. In a few minutes the cluster is stable again.

Additionally it has given us the ability to re-deploy k8s pods without having to make any changes to them. Previously we had to make a new release which I believe was due to how the endpoint cached the address. 

At the time we haven't been able to test out @dennis-tra solution :( And it might take a while since this is in production.

If you guys go ahead with that change just mention me or @Strum355 so that we're aware 👍

rogeralsing:
I do like the idea here with the connectionDecider.
This opens up for additional fixes/patches w/o having to wait for the framework codebase to be updated. 👍

rogeralsing:
btw. what kind of workloads are you all running on the clusters?
And how big are the clusters?

dennis-tra:
To spice up the discussion even further I'm proposing a third solution #350. See the details over there :)

@all how are you actually testing this problem? I'm starting consul in dev mode and two nodes that register there. Then I'm just killing one node without letting it deregister from consul. Now I observe the behaviour of the other node (log message or debugger).

Strum355:
@dennis-tra i'd update the TTLs of the consul service in iirc `consul.go` to give me a bigger window to cause the issue in and also to allow breakpoints without the still-alive node to be marked as critical

@rogeralsing we use this for an elasticsearch backed caching system deployed in k8s, an actor per user search condition. At peak times we can reach up to 1 million requests per hour in a 2 node cluster, average of around 400-500k rph. 

We're at about 700MB RAM usage total between the two, 250 on one, 450 on the other (not sure why the split of this ratio)

rogeralsing:
Emoji Vote, where are we at?:

#350 ❤️ 
#344 🚀 
#334 👀

otherview:
This *might* be connected with KV values. 

Particularly with KV pairs that belong to not-alive members that were not able to clean up.
It doesn't matter if the cluster is able to invalidate connections because when the new members are fetched from consul, it also gets old stale addresses.

This is something that we're actively seeing in our cluster :(

I've created a PR with a bit more detailed explanations around this here https://github.com/AsynkronIT/protoactor-go/pull/359

otherview:
Hi guys, I worked with @Strum355 on this. 
We can close this one out - We ended up forking and using the solution in https://github.com/AsynkronIT/protoactor-go/pull/369 and it's working for us 👍 

The EndpointWriter actor now shuts down when it can't connect instead of panicking and restarting itself.
Cheers! 😃

Issue #87, opened 2017-02-02
Scheduling messages


rogeralsing:
In Akka, there is the `ActorSystem.Scheduler`
In Proto.Actor, I would rather like to see some Scheduler interface, where we can provide different kinds of implementations.
e.g. in mem, cron job based etc.

Users can then either inject the scheduler they want into their actors, or, just have a package wide instance of whatever scheduler they want to use.

If we do this as part of the Proto.Actor library, I would like to see this as a separate package to keep the API surface as minimal as possible in the core actor package.

bryanpkc:
I wrote a `Scheduler` interface for protoactor-go which is quite similar to the `AbstractScheduler` interface in Akka. I am curious what your thoughts are on a "cron job-based" scheduler. I know how cron works but I am not seeing how the interface (with methods like `schedule` and `scheduleOnce`) works with a cron table. Can you point me to an Akka example?

rogeralsing:
In .NET we decided to go for a more opt-in model:
https://github.com/AsynkronIT/protoactor-dotnet/tree/dev/src/Schedulers/Proto.Scheduler.SimpleScheduler

It is an external package which can be evolved separately and do not require any internal changes to the core framework.

nikoszoisse:
what about the go part? IS there a scheduler?

PR #341, opened 2019-07-19
Timer-based scheduler for delayed message delivery

stuartcarnie:


rogeralsing:
Do we need to account for sending context status?
e.g. cancel sends that are scheduled after the sender context dies?

Maybe that would be weird, but at least worth discussing

stuartcarnie:
@rogeralsing I left this in draft, as I did feel there were some things to discuss around a scheduler. It seems useful as-is; however, is this the right approach?

rogeralsing:
For other readers: https://github.com/AsynkronIT/protoactor-go/issues/87

We do have this in .NET and you have used the same API here so I think we are good to go.

@AsynkronIT/developers thougths?

stuartcarnie:
@rogeralsing I did based my implementation on the .NET one for that reason ;-)

rogeralsing:
Merging this! :)

PR #344, opened 2019-08-05
Fixed panic recovery loop #2

dennis-tra:
Another approach to the problem outlined in PR #334:

We're relying on the cluster provider (consul in our case) to reflect a consistent state of the cluster topology. If a node is still part of the cluster (according to consul) we accept the panic/recovery loop, because it could be that it is just temporarily not available. As soon as the cluster provider says that the node has left the cluster the connection attempts are stopped by the `connectionDecider` - a new field on the `remoteConfig`.

In our case the `connectionDecider` checks if the address we're trying to connect to is still part of the cluster. If it is not we stop gracefully.
```go
remotingOption = append(remotingOption, remote.WithConnectionDecider(func(address string) bool {
	adrList := cluster.GetMemberAddresses()
	for _, v := range adrList {
		if v == address {
			return true
		}
	}
	log.Infof("Address %s is no known member. Deciding not to connect", address)
	return false
}))
```

Remarks: 
I would prefer if this logic didn't need to be put in from the outside but rather be handled internally. But we can't have a dependency from the `remote` to the `cluster` package -- This could be solved when the default `ClusterConfig` would have the `connectionDecider` `RemotingOption` already attached to it.

### DIFF
--- a/cluster/member_list.go
+++ b/cluster/member_list.go
@@ -106,6 +106,8 @@ func (ml *memberListValue) updateClusterTopology(m interface{}) {
 	}
 }
 
+// updateAndNotify updates the member strategy and notifies all listeners. This function may only be called with an
+// read lock on the event stream.
 func (ml *memberListValue) updateAndNotify(new *MemberStatus, old *MemberStatus) {
 	if new == nil && old == nil {
 		// ignore, not possible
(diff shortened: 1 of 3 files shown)


cluster/cluster.go
```diff
@@ -122,7 +122,23 @@ func GetMemberPIDs(kind string) actor.PIDSet {
 	return pids
 }
 
-// RemoveCache at PidCache
+// GetMemberAddresses returns the list of addresses of all currently known cluster members.
+func GetMemberAddresses() []string {
```
Strum355: Im curious why this function is here if its not being used anywhere else in the PR

dennis-tra: Good catch, this is used inside the `connectionDecider` function. I forgot to put it in here - but you can find it's usage in your PR #334. Here the excerpt:

```go
remotingOption = append(remotingOption, remote.WithConnectionDecider(func(address string) bool {
	adrList := cluster.GetMemberAddresses()
	for _, v := range adrList {
		if v == address {
			return true
		}
	}
	log.Infof("Address %s is no known member. Deciding not to connect", address)
	return false
}))
```

I'll update the PR

Strum355:
We're going to tryout this fix right now and will report back :)

dennis-tra:
Oh cool 👍 yeah let me know! :)

dennis-tra:
Oh hang on, rather try the other one :D #350 that's the one I prefer

Strum355:
Trying to figure out how to repro the issue we're having first :D this one was easier to add quickly to the code so it got preference ;)

PR #347, opened 2019-08-07
Fix: TestMpscQueueConsistency failure

### LINKED ISSUE #345: fix TestMpscQueueConsistency failure with different cpu number
When running TestMpscAQueueConsistency UT in our machine, it failed. We found the reason is that line 65 is changed to `c = runtime.NumCPU() / 2`. The cpu number of our machine is 48, then `c` is 24, which make `max` is not divisible by `c`. Therefore, line 88 will never be reachable and UT will finally fail. @stuartcarnie

yangdiangzb:
Close #345 
@stuartcarnie @PotterDai @rogeralsing


rogeralsing:
Thanks for the fix!

PR #350, opened 2019-08-08
Fix endpoint writer panic recovery loop #3

dennis-tra:
Proposal number 3 to fix the endpoint writer panic loop.

Infos:
* the actor does not `panic` anymore if a connection attempt has failed
* instead it sets its receive timeout and tries to connect again after an exponential backoff
   * the values I chose for the backoff are arbitrary
* the `*actor.ReceiveTimeout` and `*EndpointTerminated` messages ended up in a batch of `[]interface{}` messages and weren't handled by the `Receive` method of the actor directly (only indirectly in the `sendEnvelope` method).
* I added a tiny (100 entries) priority mailbox which gets drained before the user message mailbox and can only contain `*actor.ReceiveTimeout` and `*EndpointTerminated` messages
   * 100 entries is an arbitrary amount and actually I believe already 2 should suffice
* I split the endpoint writer in two behaviors (`Connecting` + `Connected`)
* both behaviors handle `*EndpointTerminated`, so if a member leaves the cluster the actor stops in every case
* The exponential backoff can be interrupted by a new batch of user messages. In this case it tries to reconnect. 
   * If the connection is successfully established the messages are sent by calling `Receive` in `endpoint_writer.go` L177. 
   * If connection establishment was unsuccessful the messages are dropped - The same happens in @Strum355s solution (if I see it right)
  * this does not happen in the current panic/restart loop implementation because the actor only gets to handle system messages - as soon as the actor is connected it drains the user message mailbox

@rogeralsing - I believe the concept of a `connectionDecider` can still be added, but may not be necessary to fix this particular problem

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,6 +17,7 @@ require (
 	github.com/hashicorp/go-cleanhttp v0.5.0 // indirect
 	github.com/hashicorp/go-rootcerts v1.0.0 // indirect
 	github.com/hashicorp/serf v0.8.2 // indirect
+	github.com/jpillora/backoff v0.0.0-20180909062703-3050d21c67d7
 	github.com/mitchellh/mapstructure v1.1.2 // indirect
 	github.com/opentracing/opentracing-go v1.0.2
 	github.com/orcaman/concurrent-map v0.0.0-20190107190726-7ed82d9cb717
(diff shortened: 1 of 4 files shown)


Strum355:
Im assuming this handles the case of where a node leaves the cluster permanently?

dennis-tra:
Hey @Strum355, the node could also be just temporarily unavailable. The endpoint writer still tries to reconnect with an exponential backoff until the node is available again or consul says the node is not part of the cluster anymore.

otherview:
Hey @dennis-tra I addressed this issue in https://github.com/AsynkronIT/protoactor-go/pull/369 if you want to have a look! 😃

PR #352, opened 2019-08-15
Fix nil pointer exception when accessing connections

dennis-tra:
When calling `remote.Shutdown(true)` the function `stopEndpointManager()` sets the `connections` map of the endpoint manager [to nil](https://github.com/AsynkronIT/protoactor-go/blob/630edf9ef64d4b178d0f53ec3bd5ebb0674d207d/remote/endpoint_manager.go#L62). Then we're waiting until the grpc server has shut down. If there is an attempt to send a message to a remote actor in this time interval (between setting the connections map to nil until the server has shut down) we get a nil pointer exception [here](https://github.com/AsynkronIT/protoactor-go/blob/630edf9ef64d4b178d0f53ec3bd5ebb0674d207d/remote/endpoint_manager.go#L101). Our "stacktrace" just shows this though:
```
sync.(*Map).Load:103
```

If we let the remote address resolver return false when the connections map is nil the `ProcessRegistry` will use the [deadletter process as a fallback](https://github.com/AsynkronIT/protoactor-go/blob/630edf9ef64d4b178d0f53ec3bd5ebb0674d207d/actor/process_registry.go#L87). Which is appropriate in my opinion.

_This not yet tested by me_

### DIFF
--- a/remote/endpoint_manager.go
+++ b/remote/endpoint_manager.go
@@ -27,6 +27,7 @@ type endpointManagerValue struct {
 	config             *remoteConfig
 	endpointSupervisor *actor.PID
 	endpointSub        *eventstream.Subscription
+	stopped            bool
 }
 
 func startEndpointManager(config *remoteConfig) {
@@ -60,7 +61,8 @@ func stopEndpointManager() {
 	rootContext.StopFuture(endpointManager.endpointSupervisor).Wait()
 	endpointManager.endpointSub = nil
 	endpointManager.connections = nil
-	plog.Debug("Stopped EndpointManager")
+	endpointManager.stopped = true
+	plog.Info("Stopped EndpointManager")
 }
 
 func (em *endpointManagerValue) endpointEvent(evn interface{}) {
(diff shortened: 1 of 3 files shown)


rogeralsing:
It looks like it does what it should, but I'm not 100% convinced this is where the error should be handled.

cc @PotterDai @stuartcarnie @AsynkronIT/developers thoughts?

dennis-tra:
While the first draft certainly worked if a new connection was established in that aforementioned short time frame, it did **not** work when a message was sent to an already existing remote PID that had its process field (`p`) already set - the `remoteHandler` function is bypassed in that case. 

I updated the PR, so that before every attempt for sending a message the state of the endpointmanager is checked. If it is in the stopped state it publishes a deadletter event. 

Now I'm not completely happy with the solution. Open for improvements @rogeralsing :)

nightwolfz:
So it has been a year.... anyone picking this up?

cupen:
> So it has been a year.... anyone picking this up?

It's fixed.

Issue #348, opened 2019-08-07
Way forward


rogeralsing:
As the Go version of Proto.Actor has the most traction, let's keep the discussion here.

How can we push the final bits in place to make this go v1 ?

What if we set up a Patreon. https://www.patreon.com/user?u=22991983
I'm not sure how likely it is that anyone actually donates to this kind of efforts.

But I'm thinking it could be used to motivate the documentation effort.
Be it external or internal devs on the project.

eg. X $ from the funds per merged page of documentation to the author.
Maybe that is just naive on my part, but coding is fun, documenting is not.
And everyone here is doing this on a spare time basis.

Is it even worth considering/evaluating?

Are there any other examples, use cases etc. that needs to be added?

dennis-tra:
> How can we push the final bits in place to make this go v1 ?

Here is my quick brainstorm:

#### Documentation:
I think this has the highest priority. I couldn't find the source for the [proto.actor](http://proto.actor/) website. Is it open for contributions?

In my opinion the first task would be to identify what's missing and what needs an update. Maybe another issue where we can build a list? Or here?

#### Test coverage
I think core packages (like the `remote` package) could benefit from an increased coverage for a stable v1 and subsequent v1.x.x releases. I'm not a friend of a policy that says at least XX % lines of code need to be tested, so it's hard to say when it suffices. I believe because of the asynchronous nature of this project it is really hard to test the functionality thoroughly though...

_Btw: The coverage badge says `unknown` at the moment._

#### Funding
I don't have any experience with OSS funding and I also don't know how likely it is for anyone to donate. But if there is no monetisation channel in place those few who would want to donate can't. So I like the idea of having Patreon in place.

rogeralsing:
The live site is generated from the content in this repo:
https://github.com/AsynkronIT/gam-web

"gam" predates Proto.Actor and was "Go Actor Model" in the very beginning :-)

dennis-tra:
Ah okay, I didn't look into the `gam-web` repo :)

What do you think of a GitHub project board that lists all issues relevant for a 1.0 release based on a tag like `v1.0`? Based on the discussion here I could create a first batch of issues and everyone who wants could throw in their own ones. If an issue is relevant or not can be discussed on that particular ticket.

With that being said, I just had a look through the different pages and here is what I noticed:

Missing pages on the top level [http://proto.actor/docs/](http://proto.actor/docs/):

1. Remote - Dead link
2. Persistance - New
3. Plugins - New
[...]

PR #357, opened 2019-09-14
Add coveralls integration settings

oklahomer:
Hi, it's my first p-r for this project. I'm a big fan of this project and created several sample repositories and blog posts including [oklahomer/protoactor-go-sender-example](https://github.com/oklahomer/protoactor-go-sender-example).

I've read #348 and mostly agreed with what is being discussed, especially about the importance of tests. However, I found out that .travis.yml does not declare coveralls.io integration even though [coveralls.io setting](https://coveralls.io/github/AsynkronIT/protoactor-go?branch=dev) is half done. This p-r tries to finish some additional .travis.yml settings including target Golang version declaration and coveralls integration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,24 @@
 language: go
+sudo: false
+
+env:
+  - GO111MODULE=on
+
+git:
+  depth: 1
+
+go:
+  - 1.12.x
+  - 1.13.x
+  - tip
+
+before_script:
+  - go get github.com/mattn/goveralls
+
 script:
-  - env GO111MODULE=on go test ./... -short
+  - go test -v -race -short -coverprofile=coverage.out ./...
+  - $GOPATH/bin/goveralls -coverprofile=coverage.out -service=travis-ci
+
+matrix:
+  allow_failures:
+    - go: tip


.travis.yml
```diff
@@ -1,3 +1,24 @@
...
+
+env:
+  - GO111MODULE=on
+
+git:
+  depth: 1
+
+go:
+  - 1.12.x
+  - 1.13.x
```
oklahomer, line 12: Wasn't so sure what versions to support, I declared the latest version and one older one.

.travis.yml
```diff
@@ -1,3 +1,24 @@
...
+  - tip
+
+before_script:
+  - go get github.com/mattn/goveralls
+
 script:
-  - env GO111MODULE=on go test ./... -short
+  - go test -v -race -short -coverprofile=coverage.out ./...
```
oklahomer, line 19: Tests can actually run by `goveralls`, but `-short` option is not currently supported so I used plain `go test` at this point. In my opinion, it would be nice if we can remove `-short` from .travis.yml so all integration tests including `consul` can be tested on CI. This could be addressed in another p-r or issue, I suppose.

.travis.yml
```diff
@@ -1,3 +1,24 @@
 language: go
+sudo: false
+
+env:
+  - GO111MODULE=on
+
+git:
+  depth: 1
```
oklahomer, line 8: Older versions of commits are not subject to test or integrate so I set `1`.

rogeralsing:
Many thanks for your contribution!

PR #358, opened 2019-09-28
Add tests for remote actors

oklahomer:
Following #357, this is a trial to have better test coverage. Since its aim is to make sure the current logic works as expected and to prevent having undesirable side effects in future modification, its modification does not contain many changes to the current codebase. However, as parts of refactoring, this work contains some minor changes introduced in a86b920 and 
5df75ed.
Being not sure if this course of action is welcomed by all collaborators, this only contains some tests on core functions as a trial. Please feel free to reject if another approach is preferred. I hope this supports #348 to release v1 in the near future.

### DIFF
--- a/remote/activator_actor.go
+++ b/remote/activator_actor.go
@@ -77,11 +77,7 @@ func Spawn(address, kind string, timeout time.Duration) (*ActorPidResponse, erro
 
 // SpawnNamed spawns a named remote actor of a given type at a given address
 func SpawnNamed(address, name, kind string, timeout time.Duration) (*ActorPidResponse, error) {
-	activator := ActivatorForAddress(address)
-	res, err := rootContext.RequestFuture(activator, &ActorPidRequest{
-		Name: name,
-		Kind: kind,
-	}, timeout).Result()
+	res, err := SpawnFuture(address, name, kind, timeout).Result()
 	if err != nil {
 		return nil, err
 	}
--- a/remote/common_test.go
+++ b/remote/common_test.go
@@ -1,0 +1,202 @@
+package remote
+
+import (
+	"fmt"
+	"github.com/AsynkronIT/protoactor-go/actor"
+	"time"
+
+	"github.com/stretchr/testify/mock"
+)
+
+// mockContext
+type mockContext struct {
+	mock.Mock
+}
+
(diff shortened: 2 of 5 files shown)


remote/activator_actor.go
```diff
@@ -77,11 +77,7 @@ func Spawn(address, kind string, timeout time.Duration) (*ActorPidResponse, erro
 
 // SpawnNamed spawns a named remote actor of a given type at a given address
 func SpawnNamed(address, name, kind string, timeout time.Duration) (*ActorPidResponse, error) {
-	activator := ActivatorForAddress(address)
-	res, err := rootContext.RequestFuture(activator, &ActorPidRequest{
-		Name: name,
-		Kind: kind,
-	}, timeout).Result()
+	res, err := SpawnFuture(address, name, kind, timeout).Result()
```
oklahomer, line 80: Unified common logic in a86b920. This should not include any behavioral change.

remote/common_test.go
```diff
@@ -0,0 +1,202 @@
+package remote
```
oklahomer, line 1: Code is borrowed from actor/common_test.go and router/common_test.go to follow the same testing strategy.

remote/server.go
```diff
@@ -23,7 +22,7 @@ var rootContext = actor.EmptyRootContext
 
 // Start the remote server
 func Start(address string, options ...RemotingOption) {
-	grpclog.SetLogger(slog.New(ioutil.Discard, "", 0))
+	grpclog.SetLoggerV2(grpclog.NewLoggerV2(ioutil.Discard, ioutil.Discard, ioutil.Discard))
```
oklahomer, line 25: 5df75ed replaces the usage of the deprecated method.
Since the previous logger used `ioutil.Discard`, replacing code also uses `ioutil.Discard` for all levels.
If this is preferable to set separate `io.Writer` for each log level, that can be done in the following pull request.

cupen:
nice work.

oklahomer:
Could I have your opinion about this? @rogeralsing @PotterDai

potterdai:
@oklahomer Thank you very much for your work and really sorry for the delay! I will merge it right the way! =D

PR #359, opened 2019-10-14
Removing the KVs usage from consul

otherview:
Hi !

We've been tackling an issue where the grains try to connect to non-existing instances of silos.
This seems to work - would like to have a review from you guys, I might be missing something :)

# Breakdown of the problem
- Multiple instances of silos connect lets say addresses/IDs: (X,Y,Z)
- Instances register with consul as service - MyCluster
- Instances ask consul to store some more data in the form of K/V pairs
- Instances list the keys and from those they assert which silos exist
- MyCluster is now 1 service with 3 instances (X, Y, Z) and 3 K/V pairs
(problem)
- We kill all 3 instances and restart them on different addresses (A,B,C)
- Instances register with consul as a service - MyCluster
- Instances ask consul to store some more data in the form of K/V pairs
- Instances list the keys and from those they assert which silos exist
Now this brings back 6 KV (A,B,C,X,Y,Z) because there was never a cleanup of the KV.
- The problem then continues on because there is no instance that can clean up the KV values of X,Y,Z

# Fix
Stop using the KV.
Instead use the health endpoint and get the address from the Service, leaving Consul to manage available ones.

# Caveat 
- I'm not sure what we want from K/V pairs
- Not sure of the purpose of the `memberStatusVal` here: https://github.com/AsynkronIT/protoactor-go/blob/e7cc7b4f4be0bd4c6343a1a2ca39854db8dcdd65/cluster/consul/consul_provider.go#L255

Thanks! :)

### DIFF
--- a/cluster/consul/consul_provider.go
+++ b/cluster/consul/consul_provider.go
@@ -63,11 +63,6 @@ func (p *ConsulProvider) RegisterMember(clusterName string, address string, port
 		return err
 	}
 
-	err = p.registerMember()
-	if err != nil {
-		return err
-	}
-
 	// IMPORTANT: do these ops sync directly after registering.
 	// this will ensure that the local node sees its own information upon startup.
 
@@ -86,11 +81,6 @@ func (p *ConsulProvider) RegisterMember(clusterName string, address string, port
 
 func (p *ConsulProvider) DeregisterMember() error {
 	err := p.deregisterService()
-	if err != nil {
-		fmt.Println(err)
-		return err
-	}
-	err = p.deregisterMember()
 	if err != nil {
 		fmt.Println(err)
 		return err
@@ -138,13 +128,6 @@ func (p *ConsulProvider) UpdateTTL() {
 				continue
 			}
 
-			err = p.registerMember()
-			if err != nil {
-				log.Println("[CLUSTER] [CONSUL] Error reregistering member", err)
-				time.Sleep(p.refreshTTL)
-				continue
-			}
-
 			log.Println("[CLUSTER] [CONSUL] Reregistered service in consul")
 			time.Sleep(p.refreshTTL)
 		}
@@ -156,12 +139,7 @@ func (p *ConsulProvider) UpdateMemberStatusValue(statusValue cluster.MemberStatu
 	if p.statusValue == nil {
 		return nil
 	}
-	kvKey := fmt.Sprintf("%v/%v:%v/StatusValue", p.clusterName, p.address, p.port)
-	_, err := p.client.KV().Put(&api.KVPair{
-		Key:   kvKey,
-		Value: p.statusValueSerializer.ToValueBytes(p.statusValue), // currently, just a semi unique id for this member
-	}, &api.WriteOptions{})
-	return err
+	return nil
 }
 
 func (p *ConsulProvider) blockingUpdateTTL() error {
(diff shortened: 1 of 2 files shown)


Strum355:
What's the cenas file for?

otherview:
Thks @Strum355  :)

otherview:
Bumpity bump 🍨

potterdai:
@otherview Thank you very much for the PR! Sorry for late merge... =P

potterdai:
@otherview Sorry was merging a bit too fast...I reverted the PR just for now! = )

@rogeralsing This PR removes all KV usage in Consul. The good thing is that it seems more stable than relying on Consul's KV, however the MemberStatusValue will not function in this case.

@otherview The reason MemberStatusValue was introduced is to let user define some custom values for different members, for example in this https://github.com/AsynkronIT/protoactor-go/blob/dev/cluster/weighted/weighted_member_status.go , the weight value is stored in Consul's KV so that grains can be spawned according to different weight defined for each member(Silo), rather than be spawned evenly across all Silos.

otherview:
Thanks for the review @PotterDai 😃 !
🤔 That makes sense.

I do have a few questions about it.. Is there an example of weighted members usages?
Would make it easier to have a starting point.
I've created a simple test with 
```
consulServiceID := "cluster-broadcast-test"
clusterConfig := cluster.NewClusterConfig(consulServiceID, fmt.Sprintf("127.0.0.1:%v", port), cp)
clusterConfig.MemberStrategyBuilder = weighted.NewWeightedMemberStrategy
clusterConfig.MemberStatusValueSerializer = &weighted.WeightedMemberStatusValueSerializer{}
cluster.StartWithConfig(clusterConfig)
```

Should I override the `WeightedMemberStatusValueSerializer` or something ? Not sure where/how this uses the weight to make decisions. Would appreciate pointers 😄 

I've also been looking into Service metadata, it's a good candidate for replacing the KV pairs as it's tied to the service/node being alive or not.

Really want to push this one ahead as it's a problem in our prod env. :) 
But ultimately KV's are a pain in the arse as they won't clear away even if the cluster is all down.

Also I think is connected with this pr https://github.com/AsynkronIT/protoactor-go/pull/334
The cluster keeps trying to connect to the same entity and failing because it's getting an old KV from consul that doesn't exist anymore.

potterdai:
@otherview Sorry for the confusion!
The usage of WeightedMemberStatusValue can be found here
https://github.com/AsynkronIT/protoactor-go/blob/dev/cluster/weighted/weighted_round_robin.go#L47
This is just an example, which will spawn grain unevenly across different silos.
Does Consul have its embedded feature that can achieve similar functionality? The logic behind it is actually very simple, that each member can have an assignable value, and this value will sync across all members.

otherview:
No problem ! :)

I'll give it a go with the weighted round robin !

Don't quote me on this, it's just my understanding from using it.
When we register a Service we're actually registering an instance of a service that live in a node.
Registering `MyCluster` as a service creates an instance of MyCluster@address
image

I've started two nodes on 2 different addresses 
image

And then I added Metadata to their nodes
image

Which we can do by adding it on the `Meta` field here: https://github.com/AsynkronIT/protoactor-go/blob/0beb1f1a977cccc51bf3040d4a1b83f7ea7ed457/cluster/consul/consul_provider.go#L151-L160


In the cluster package we have `registerMember` which we basically use to register the KV pair, maybe we can move that logic to the `registerService` ?

The big upside is that we tied the metadata to the service@node and it will not outlive the service.
That's just my thoughts at the moment, what do you think ? :)

potterdai:
@otherview I just did a quick research on Consul's metadata feature.
It actually works! Let's switch to that. Could you create a new PR for this since I just reverted this PR and it seems Github doesn't re-open after reverting. Then we can just move MemberStatusValue into Consul's meta = )
Thank you very much for your effort! =D

potterdai:
@otherview Sorry for my delay! I will make a new PR to fix this weekend. = )

otherview:
Hey @PotterDai ! That's totally cool. 
I've been pulled to something else, I'm trying to sneak some time into writing that PR but no luck so far. :(

Finding that round robin code tricky as hell 😅

PR #369, opened 2020-01-18
Endpoint actor now doesn't crashloop

otherview:
Removing the panic as it would only keep the Endpoint writer trying to connect to an address that did not exist anymore.
Instead it returns an error and stops the actor - instead of panic + restart forever.

This has been a major bug in our cluster that would require a total restart of the cluster node. 
Very happy to have caught it 😃

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -1171,6 +1171,7 @@ github.com/opentracing-contrib/go-observer v0.0.0-20170622124052-a52f23424492/go
 github.com/opentracing/basictracer-go v1.0.0/go.mod h1:QfBfYuafItcjQuMwinw9GhYKwFXS9KnPs5lxoYwgW74=
 github.com/opentracing/opentracing-go v1.0.2 h1:3jA2P6O1F9UOrWVpwrIo17pu01KWvNWg4X946/Y5Zwg=
 github.com/opentracing/opentracing-go v1.0.2/go.mod h1:UkNAQd3GIcIGf0SeVgPpRdFStlNbqXla1AfSYxPUl2o=
+github.com/opentracing/opentracing-go v1.1.0 h1:pWlfV3Bxv7k65HYwkikxat0+s3pV4bsqf19k25Ur8rU=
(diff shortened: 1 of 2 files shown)


dennis-tra:
Hey @otherview,

nice that you caught the bug :) Your solution certainly works  :+1:

However I think there is another question to be asked when you state:
> [...] trying to connect to an address that did not exist anymore.

Just by trying to connect to it we can't be sure if it is just was a network issue or the remote node is gone permanently. So, who gets to decide whether an address does not exist anymore? 

In my opinion the cluster provider should have the sole authority about who is part of the cluster and who is not. By stopping to connect to the remote node the `EndpointWriter` here decides that on its own. Instead it should continue trying until the cluster provider says that the remote node is not part of the cluster anymore.

This also seems to be the intention of the current `dev` branch implementation. 

However, there the `EndpointWriter` never gets to process the `EndpointTerminated` messages (**user** messages) which would lead to a stop because the actor is busy processing the `*actor.Started` and `*actor.Restarting` messages (**system** messages) from the `panic`. This leads to the endless restart loop.

The PR #350 takes these thoughts into consideration.

otherview:
Valid points! And it's very true, this is a blunt approach.

I suppose the former approach is too soft ?
When we're dealing with millions of messages, waiting for an event from the cluster doesn't seem to cut it - it can be delayed or not arrive at all.

We're using Automanaged clustering where the cluster topology in every node gets updated every 2 seconds and it would still go into and unrecoverable state :(

I'd prefer to kill endpoints that can't connect and let them retry by normally re-instantiating themselfs with the refreshed addresses. 

Even if it's just a connection error it should recover faster than the potential 2 seconds that (in the best case scenario) would take the cluster to update its topology.

I'm happy with the results but I'd be more than happy to compromise in something in between :)

cupen:
LGTM. ;p

rogeralsing:
a bit of merge conflicts here

rogeralsing:
Highjacking this PR to add contributors to the readme.

@all-contributors please add @rogeralsing for code

rogeralsing:
@all-contributors please add @rogeralsing for code

Issue #148, opened 2017-07-25
cluster: manage cluster membership by using gossip based protocol


RussellLuo:
Currently, `protoactor-go` only supports cluster management via Consul. While this looks good and is feasible, but it forces a strong dependence on Consul at the same time.

As far as I know:
- Akka cluster is [using gossip protocol](http://doc.akka.io/docs/akka/2.5.3/scala/common/cluster.html)
- Orleans uses their own membership protocol, which is also a [gossip-like peer-to-peer membership protocol](https://dotnet.github.io/orleans/Documentation/Runtime-Implementation-Details/Cluster-Management.html#the-basic-membership-protocol)

So I think it maybe wonderful if we add a gossip based implementation of cluster provider, and also allow users create their own implementation.

Here I have already implemented a [gossip_provider](https://github.com/RussellLuo/protoactor-go/tree/feature-add-gossip-provider/cluster/gossip), and this is [the example](https://github.com/RussellLuo/protoactor-go/tree/feature-add-gossip-provider/examples/gossipcluster). I can submit a pull request if it's the way `protoactor-go` want to support.

cpx86:
More providers is nice - it's always a question of what we should support as part of the core library though, and what should be treated more as "extensions" for lack of a better word.

For example one option could be to support Consul as the default/bundled provider, and keep other providers in their own repos, either under this organisation or some other.

Personally I'm not 100% decided to be honest. @AsynkronIT/developers what do you others think? :)

RussellLuo:
Personally, I prefer the philosophy that does not force a strong dependence on any particular technology , which is also part of [the design rationale of Orleans](https://dotnet.github.io/orleans/Documentation/Runtime-Implementation-Details/Cluster-Management.html#design-rationale):

> as part of Orleans's philosophy, we do not want to force a strong dependence on any particular technology, but rather have a flexible design where different components can be easily switched with different implementations.

So I wish the gossip based provider can become the default provider, which is also the default/core membership protocol like Akka and Orleans have adopted.

rogeralsing:
One of the main tennants of Proto.Actor is to not reinvent the wheel, if there are tools that solve a problem, then use those tools.
Cluster membership is tricky to get right, it's a bug prone process.

We also have the aspect of being cross platform, Go Actors work with C# Actors etc.
Having our own cluster membership implementation would require us to have the exact same implementation across languages.
Any little difference could cause a cluster to fail.

[...]

PR #370, opened 2020-01-18
Adding an Automanaged Cluster package

otherview:
This is a package that can be used for the cluster management instead of using Consul.

In short: 
- There's a hardcoded fixed list of members
- Each member has that list
- Each member pings that list and updates the cluster topology
- Each member has an open http port where he reports his information (alive, kinds, etc)

Simple enough for what we use - we have a k8s cluster with 10 pods where we were not getting much from consul.
Additionally there are some issues with consul that would inevitably hit the protoactor cluster.
I decided to get rid of consul and limit our exposure to issues.

With this packaged and the https://github.com/AsynkronIT/protoactor-go/pull/369 we now have more reliant cluster.
Before, if a pod would restart (change IP) the cluster would go bananas on errors and some actors would never respond again (PR https://github.com/AsynkronIT/protoactor-go/pull/369 ). 
Additionally if the consul pods would have problems electing a leader or *any* kind of problem the cluster would go bananas again. Reducing the risk seems to be the right choice - for now anyways 😅 

Happy to discuss this further! 😃

### DIFF
--- a/cluster/automanaged/automanaged_provider.go
+++ b/cluster/automanaged/automanaged_provider.go
@@ -1,0 +1,390 @@
+package automanaged
+
+import (
+	"encoding/json"
+	"fmt"
+	"net"
+	"net/http"
+	"sync"
+	"time"
+
+	"golang.org/x/net/context"
+
+	"github.com/AsynkronIT/protoactor-go/cluster"
+	"github.com/AsynkronIT/protoactor-go/eventstream"
+	"github.com/AsynkronIT/protoactor-go/log"
+	"github.com/labstack/echo"
+	"golang.org/x/sync/errgroup"
+)
+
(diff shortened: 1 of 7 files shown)


otherview:
@rogeralsing Is this PR something that goes against this ? 🤔 Or is there space to maneuver ?
  

> **Build on existing technologies** - There are already a lot of great tech for e.g. networking and clustering, build on those.
> e.g. gRPC streams for networking, Consul.IO for clustering.

cupen:
It seems like this? :smile:
https://github.com/AsynkronIT/protoactor-go/issues/148

otherview:
Closing this one in preference to https://github.com/AsynkronIT/protoactor-go/pull/394

PR #371, opened 2020-01-21
Fix deadlock when (un)subscribing from/to the eventstream

dennis-tra:
This PR introduces the method `eventstream.PublishUnsafe` which does not acquire the read lock on the eventstream.

When a `ClusterTopologyEvent` is published on the event stream the emitter of this event has acquired a read lock to the event stream. As an effect of this event another call to `eventstream.Publish` is **synchronously** made in `updateAndNotify` further down the chain. So a second read lock gets acquired (`eventstream/eventstream.goL65`)

When right in between both of those two read locks another write lock wants to be made we end up in a dead lock. The order of events:
* read lock gets acquired
* write lock -> waits until first read lock is released
* second read lock -> waits until the write lock is released

That the second read lock waits for the write lock is an implementation detail of the `sync.RWMutex` `RLock` method:
```go
// RLock locks rw for reading.
//
// It should not be used for recursive read locking; a blocked Lock
// call excludes new readers from acquiring the lock. See the
// documentation on the RWMutex type.
func (rw *RWMutex) RLock() {
	if race.Enabled {
		_ = rw.w.state
		race.Disable()
	}
	if atomic.AddInt32(&rw.readerCount, 1) < 0 {
		// A writer is pending, wait for it.          <-------
		runtime_SemacquireMutex(&rw.readerSem, false, 0)
	}
	if race.Enabled {
		race.Enable()
		race.Acquire(unsafe.Pointer(&rw.readerSem))
	}
}
```
Source: https://golang.org/src/sync/rwmutex.go

It is safe to use the `PublishUnsafe` method in `updateAndNotify` because it is synchronously made as a result of the `ClusterTopologyEvent`.

The given test proves that there was a deadlock. When the `PublishUnsafe` methods are replaced with the regular `Publish` methods we run into a timeout.

### DIFF
--- a/cluster/member_list.go
+++ b/cluster/member_list.go
@@ -106,6 +106,8 @@ func (ml *memberListValue) updateClusterTopology(m interface{}) {
 	}
 }
 
+// updateAndNotify updates the member strategy and notifies all listeners. This function may only be called with an
+// read lock on the event stream.
 func (ml *memberListValue) updateAndNotify(new *MemberStatus, old *MemberStatus) {
 	if new == nil && old == nil {
 		// ignore, not possible
@@ -129,13 +131,13 @@ func (ml *memberListValue) updateAndNotify(new *MemberStatus, old *MemberStatus)
 			Kinds: old.Kinds,
 		}
 		left := &MemberLeftEvent{MemberMeta: meta}
-		eventstream.Publish(left)
+		eventstream.PublishUnsafe(left)
 		delete(ml.members, old.Address()) // remove this member as it has left
 
 		rt := &remote.EndpointTerminatedEvent{
 			Address: old.Address(),
 		}
-		eventstream.Publish(rt)
+		eventstream.PublishUnsafe(rt)
 
 		return
 	}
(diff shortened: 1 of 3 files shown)


potterdai:
@dennis-tra Sorry for late reply! Wow, this is a great catch! = )

PotterDai:
@dennis-tra I just double checked ProtoActor-Dotnet and it seems instead of a mutex lock, it uses the default synchronized dispatcher to manage messages(similar to actors). https://github.com/AsynkronIT/protoactor-dotnet/blob/dev/src/Proto.Actor/EventStream.cs 
This is something I'll investigate in Go as well, perhaps in the end we can get rid of the mutex lock in EventStream.

potterdai:
@dennis-tra I just made a PR, which uses a concurrent map instead of array and mutex.
https://github.com/AsynkronIT/protoactor-go/pull/376

dennis-tra:
> @dennis-tra I just made a PR, which uses a concurrent map instead of array and mutex.
> #376

I've commented on your PR :+1:

Issue #374, opened 2020-02-20
Cluster: consul provider writes to service-meta but reads from node-meta


D-Raiser:
When registering or updating a member via `RegisterMember`, `UpdateTTL` or `UpdateMemberStatusValue`, the `MemberStatusValue` is written the `StatusValue` field of the service-metadata.
In `notifyStatuses` however the `MemberStatusValue` is read from the metadata of a consul-node which leads to always passing an empty value to the `MemberStatusValueSerializer`.

D-Raiser:
write: https://github.com/AsynkronIT/protoactor-go/blob/dev/cluster/consul/consul_provider.go#L152
read: https://github.com/AsynkronIT/protoactor-go/blob/dev/cluster/consul/consul_provider.go#L193

PR #375, opened 2020-02-20
Fix empty MemberStatusValues in consul cluster provider #374

D-Raiser:
Instead of reading from the node metadata the consul cluster provider
now reads the serialized MemberStatusValue from consul's service metadata where it was already being written to.


potterdai:
@D-Raiser Thank you for the fix! Merge right the way! = )

potterdai (approved):

PR #376, opened 2020-02-21
Use cmap instead of array/mutex for Subscriptions in EventStream

potterdai:
This removes the concurrency issue mentioned in this PR https://github.com/AsynkronIT/protoactor-go/pull/371

### DIFF
--- a/eventstream/eventstream.go
+++ b/eventstream/eventstream.go
@@ -2,12 +2,17 @@ package eventstream
 
 import (
 	"sync"
+
+	"github.com/google/uuid"
+	cmap "github.com/orcaman/concurrent-map"
 )
 
 // Predicate is a function used to filter messages before being forwarded to a subscriber
 type Predicate func(evt interface{}) bool
 
-var es = &EventStream{}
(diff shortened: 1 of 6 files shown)


eventstream/eventstream.go
```diff
@@ -21,61 +26,34 @@ func Publish(event interface{}) {
...
-
-func (es *EventStream) PublishUnsafe(evt interface{}) {
-	for _, s := range es.subscriptions {
-		if s.p == nil || s.p(evt) {
-			s.fn(evt)
+	for _, key := range es.subscriptions.Keys() {
+		i, ok := es.subscriptions.Get(key)
+		if ok {
+			s, ok := i.(*Subscription)
+			if ok && (s.p == nil || s.p(evt)) {
```
dennis-tra, line 54: I'm not entirely sure about this:
Imagine `s.p == nil` evaluates to false, `WithPredicate(nil)` is called and then `s.p(evt)` is evaluated. This would result in an NPE.

Hence I don't think it is save to access `s.p` here without any locking mechanism.

Issue #377, opened 2020-02-24
Consul cluster race-condition: Service re-registration after cluster shutdown


D-Raiser:
When `Shutdown` is called on the consul provider it is possible that `DeregisterMember` resolves before `blockingUpdateTTL` resolves in the loop in `UpdateTTL`. If this happens, `blockingUpdateTTL` will return an error since the member for which it tries to update the TTL does not exist anymore. This eventually leads to `UpdateTTL` re-registering the member in consul and immediately after that stopping since the condition for the for loop is not `true` anymore (as the provider was just shutdown). As a result the member will exist in consul but be marked as `critical` since the health-check is failing.

A service re-registration can also happen when `UpdateMemberStatusValue` is called after `cluster.Shutdown`. Since `UpdateTTL` will not be running anymore at this point this leads to the service being marked as critical in consul for the duration `deregisterCritical` of the consul provider.

PR #378, opened 2020-02-24
Fix consul member re-registration after shutdown #377

D-Raiser:
When the cluster was shut down and the member deregistration resolved while the health check TTL was
updated, the member was re-registered and as a result was displayed in
consul for 1 minute with the status 'crititcal'.
A re-registration could also happen when the member's status value was
updated after the shutdown finished.

### DIFF
--- a/cluster/consul/consul_provider.go
+++ b/cluster/consul/consul_provider.go
@@ -3,11 +3,18 @@ package consul
 import (
 	"fmt"
 	"log"
+	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


D-Raiser:
The race-condition with `UpdateTTL` and `Shutdown` can be demonstrated with the Test `TestUpdateTTLDoesNotReregisterAfterShutdown` when removing `p.updateTTLWaitGroup.Wait()` in line 106 in the `consul_provider.go`

PR #384, opened 2020-06-06
move examples -> _examples

stuartcarnie:
* Clean up go.mod to remove all unnecessary dependencies from `protoactor-go`.

* Create individual go.mod files in each example, so they have their own dependencies that are only required if actually run.


stuartcarnie:
@rogeralsing this makes adding protoactor-go as a dependency much less painful

PR #380, opened 2020-02-27
Add UnboundedPriority mailbox type

cap10morgan:
Prioritizes messages into 0-7 (0 = lowest, 7 = highest) based on return value of their GetPriority method. Based on discussion in #379.

I benchmarked this using `examples/spawnbenchmark` and got these results:

existing unbounded Goring: 499999500000 in 5328 ms
priority unbounded Goring (with all messages getting default priority): 499999500000 in 7252 ms
priority unbounded Goring (with randomized priority on messages): 499999500000 in 8090 ms

PR #392, opened 2020-08-15
Add UnboundedPriority mailbox type

cap10morgan:
Prioritizes messages into 0-7 (0 = lowest, 7 = highest) based on return value of their GetPriority method.

Rebased & conflict-free version of #380.


rogeralsing:
Many thanks! 🎉

PR #368, opened 2020-01-14
Redis cluster

otherview:
Initial push for the redis package.

Currently its failing because of race conditions with the mock package

PR #393, opened 2020-08-19
Redis cluster approach + updating cluster-broadcast example

otherview:
Second push for the redis package.
Followup and the clean up from https://github.com/AsynkronIT/protoactor-go/pull/368 . ( It was very messy 😅  )

This approach uses redis as a manager of nodes that are alive.

- Each node registers itself to redis with an `expiration` time
- Each node `SCAN`s redis searching `clusterName` keys
- Each key represents an existing node (wich update the cluster using a ClusterTopologyEvent )


_examples/cluster-broadcast/node1/main.go
```diff
@@ -2,28 +2,37 @@ package main
 
 import (
 	"fmt"
-	"github.com/AsynkronIT/goconsole"
+	"time"
+
+	console "github.com/AsynkronIT/goconsole"
+	"github.com/AsynkronIT/protoactor-go/_examples/cluster-broadcast/shared"
```
cupen, line 8: @otherview  Hello. :smiley: 
This `import` may not work with `go module`.
Here is my pr for  `go module` #395 , it would breaking `GOPATH` with `vendor`.

PR #394, opened 2020-08-21
Adding automanaged cluster provider

otherview:
Following https://github.com/AsynkronIT/protoactor-go/pull/370 - cleaning up some of the git blunders + updating the tests :)

> This is a package that can be used for the cluster management instead of using Consul.
> 
> In short:
> 
> There's a hardcoded fixed list of members
> Each member has that list
> Each member pings that list and updates the cluster topology
> Each member has an open http port where he reports his information (alive, kinds, etc)
> Simple enough for what we use - we have a k8s cluster with 10 pods where we were not getting much from consul.
> Additionally there are some issues with consul that would inevitably hit the protoactor cluster.
> I decided to get rid of consul and limit our exposure to issues.
> 
> With this packaged and the #369 we now have more reliant cluster.
> Before, if a pod would restart (change IP) the cluster would go bananas on errors and some actors would never respond again (PR #369 ).
> Additionally if the consul pods would have problems electing a leader or any kind of problem the cluster would go bananas again. Reducing the risk seems to be the right choice - for now anyways 😅
> 
> Happy to discuss this further! 😃


rogeralsing:
Many thanks for this one! 
Please bear with us while we try this out before merging :-)

nightwolfz:
Looks interesting. Will this get merged?

otherview:
Just wanted to let know that we're using this in production for more than 9 months now.

EDIT: And then I hit close the PR 😅

rogeralsing:
cc @otherview any chance you could realign all this ontop of this branch?
https://github.com/AsynkronIT/protoactor-go/pull/412

I'm working on aligning all the APIs with the dotnet version. 
Having this `ActorSystem` allows us to run multiple nodes in the same process, e.g. testing cluster features and such

PR #395, opened 2020-09-01
Fix examples which work with go modules.

cupen:
* Fix `import path` of many examples, it works with  `go modules` right now.
* Add some Makefile for the multi-processes examples. 
  use `tmux` and just type `make start`, then you'll see all the processes logs in the same terminal session.
  `ctrl-b d` will exit and kill the tmux session, then all the processes will be stopped.
  
image

/cc @stuartcarnie 

BTW:  I don't know how to run `_examples/remotessl` ,  the  `make nodes` does'nt works.


rogeralsing:
Fantastic!

PR #397, opened 2020-09-27
Fix log time

cupen:
A little fix. After this `pr`, the log would works with time and prefix.

```bash
2020/09/27 11:16:25 [REMOTE] Started EndpointManager 
2020/09/27 11:16:25 [REMOTE] Starting Proto.Actor server address="127.0.0.1:8080" 
2020/09/27 11:16:25 [REMOTE] Started Activator 
2020/09/27 11:16:25 [CLUSTER] Starting Proto.Actor cluster address="127.0.0.1:8080" 
2020/09/27 11:16:25 [CLUSTER] Started kind="Hello" id="partition-Hello"
```

But there is a question, why not print local time? 
https://github.com/AsynkronIT/protoactor-go/blob/b225d28383f2318e0513dfabd22a3037f9043ffd/log/string_encoder.go#L55

/cc @PotterDai  @rogeralsing

### DIFF
--- a/log/string_encoder.go
+++ b/log/string_encoder.go
@@ -35,7 +35,7 @@ func (l *ioLogger) listenEvent() {
 }
 
 // Cheap integer to fixed-width decimal ASCII.  Give a negative width to avoid zero-padding.
-func itoa(buf bytes.Buffer, i int, wid int) {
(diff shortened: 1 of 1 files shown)


rogeralsing:
👍🏻 many thanks for the fix!

PR #419, opened 2020-11-09
Add default instance of ActorSystem and Remote.

cupen:
For most cases, It's a easier way to use one node in one ·`os/process`.


cupen:
En... maybe not. , 
My intention is make it easier to use in most cases. Anyway, this is the idiomatic way in golang. It's similar as `net/http`.
https://github.com/golang/go/blob/feccfb8adaf5a0ce93a0dafa31336ccb6f41c618/src/net/http/client.go#L108-L109
https://github.com/golang/go/blob/feccfb8adaf5a0ce93a0dafa31336ccb6f41c618/src/net/http/client.go#L441-L447

rogeralsing:
Not sure about the ActorSystem TBH, the remote bits looks totally fine.
I'm just afraid that users might mistakenly use actor.System somewhere instead of the system they intended to use. and get hard to track bugs from that.

In the dotnet version, we also have configuration for the actor system. e.g. throttling for deadletter logging. that would not be possible to port over if there is a default system.

cupen:
@rogeralsing You are right. My idea is not mature enough.

PR #420, opened 2020-11-09
Fix compiling error of testcases.

cupen:
I think the testcases of ConsulProvier need to rewrite. I fix the compiling error, but it's still failed.


rogeralsing:
OK, I will merge this in and work from there 👍🏻

PR #423, opened 2020-11-16
refactor cluster.ConsulProvider and its testcases.

cupen:
It seems there are a lot work need to to.
I ported the ClusterProvider interface from `dotnet version`. It's a bit step.

btw: I'm a etcd user. I think make a `TestSuite` for cluster would be good for ClusterProviders.


rogeralsing:
Very nice 👍🏻

cupen:
@otherview Hello :smiley: 
After this `pr`,  `AutoManagedProvider` will be broken. It hasn't testcases, so I can't refactor it, sorry for that.

PR #428, opened 2020-11-23
Add cluster.Call and refactor grain codegen

cupen:
@rogeralsing  @PotterDai @otherview   Hello guys. 
Here is some major changes for grain codegen, sorry it's not incompatible.
1. Merge methods `XX` and `XXWithOpts` into one.
2. Drop all of the methods named  `XXChan`, because Go has'nt magic for async call.
    So just use goroutine with channel, even though it's not pretty good with super-high concurrency.
3. Just use `time.Duration` for grain timeout, it's an alias of int64.
4. Add a interface `ReceiveDefault` for grain, use for some advanced custom.

I hope it would be easier to use. Feel free to tell me what you think.
Regards!


cupen:
OK. I create a new one. 
usage:
```bash
cd ./protobuf/protoc-gen-gograinv2
make install 
# make sure $HOME/go/bin in your $PATH
```
then
```bash
protoc -I=.  --gograinv2_out=. shared/*.proto
```

PR #431, opened 2020-11-27
Initial port

umitkavala:
There are some changes happened on .net version since this issue created. I just want to make sure that I'm in correct path. 
I've looked at MongoDb implementation : https://github.com/umitkavala/protoactor-dotnet/blob/09e5ec1cb99da6fb4f098415e8de2d877c3330f7/src/Proto.Cluster.Identity.MongoDb/MongoIdentityStorage.cs#L10
But we don't use IIdentityLookup interface there instead we use IdentityStorage. 

What I understand is we need : https://github.com/umitkavala/protoactor-dotnet/blob/09e5ec1cb99da6fb4f098415e8de2d877c3330f7/src/Proto.Cluster.Identity/IIdentityStorage.cs#L7

Question is : are we going to port Proto.Cluster.Identity project? because when I look at Mongodb project we don't use any class from that.

If I try to port Proto.Cluster.Identity project, as you can see I've already stuck at
```
func (i *IdentityStorageLookup) Get(clusterIdentity *cluster.ClusterIdentity) *actor.PID {
	msg := newGetPid(clusterIdentity)
	i.system.Root.Request(i.router, msg)
}
```
because i.system.Root.Request doesn't return PID. 

Could you please advise? IIdentityStorage + Proto.Cluster.Identity.MongoDb port?


rogeralsing:
>Question is : are we going to port Proto.Cluster.Identity project? 

Yes i think we should.
Identity project is an abstraction which makes it possible to reuse a lot of the underlying code and the specific implementations can then just focus on the storage parts.

cc @mhelleborg for more context

PR #433, opened 2020-11-28
Gracefull remote shutdown

Damian-P:
Hello,
As this is my first PR in go language, it's maybe not perfect :-)  but here you will find some code to gracefully handle remote shutdown. I used to same approach in the dotnet project.

When a remote instance is about to stop, the endpointReader has to send a unit message to all of the endpointWriters to stop the connections.
In grpc, the caller has to terminate the connection. And to gracefully shutdown, the grpc server has to finish all the incoming queries.

For example :
- On Node A endpointManger is stopping, it tells to all active receive methods on endpointReaders to send a Unit message.
- On Node X endpointWriter receives the unit message and publishes an EndpointTerminatedEvent event
- On Node X endpointManager handles the EndpointTerminatedEvent event and stops the endpoint actors for this address. When the endpointWriter stops it closes the connection to Node A.

### DIFF
--- a/_examples/remote-benchmark/node1/main.go
+++ b/_examples/remote-benchmark/node1/main.go
@@ -7,6 +7,7 @@ import (
 
 	"remotebenchmark/messages"
 
+	console "github.com/AsynkronIT/goconsole"
 	"github.com/AsynkronIT/protoactor-go/actor"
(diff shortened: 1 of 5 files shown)


rogeralsing:
Very nice!

PR #434, opened 2020-12-09
Fix missing publish SupervisorEvent from SupervisorStrategy handler

iwaltgen:
Some SupervisorStrategy does not publish events in the HandleFailure method.
- exponential backoff strategy
- restarting strategy

I think all supervisors should publish the same events in the same action.


rogeralsing:
👍🏻 Thank you

Issue #409, opened 2020-11-01
Port ClusterInit to Go


rogeralsing:
https://github.com/AsynkronIT/protoactor-dotnet/blob/dev/src/Proto.Cluster/Partition/PartitionPlacementActor.cs#L172

This is basically a small message with 3 fields, Cluster (the actual cluster instance), Identity and Kind.
This is useful for virtual actors to know their own name in the cluster.
Especially useful when combined with persistence, so the actor knows what state to load

cupen:
How about get cluster instance and grainID from `cluster.GrainContext` ?

cupen:
Or how about  `context.SpawnWithArgs` ?

LostLaser:
I would like to pick this up. I am a bit confused by the provided link though. I would think the fix for this issue would just be adding a new type to a proto file?

cupen:
It's done.
https://github.com/AsynkronIT/protoactor-go/blob/df2fc305778c5875a88880f2eb376a4298622f73/cluster/partition_placement_actor.go#L168-L177

PR #435, opened 2020-12-14
new cluster partition implementation.

cupen:
Be careful!! There were a lot breaking changes about cluster. 
* new partition implementation ported from `C# version`. (see `partition_identity.go`, `partition_placement.go`, `partition_kind.go`)
* new rendezvous implementation and benchmarks.
* add ClusterInit message, for #409.
* add a UML sequence diagram for describing the new partition.(see `cluster/partition.puml`).
* update grainv2 codegen.
* update some examples

btw: Add a cluster provider. I'm using etcd in my production, it works pertty fine.


cupen:
The new partition was based on `MemberStrategy`, which is the same as the old implementation. This is different to the `C# version`.
https://github.com/AsynkronIT/protoactor-dotnet/tree/dev/src/Proto.Cluster/Partition

Feel free to tell me what should be improved.
cc @rogeralsing @PotterDai @otherview

rogeralsing:
Impressive!

This is going to take some time to review as its 80+ files changed. 
cc @mhelleborg maybe you could take a look also as we are now getting the cluster_identity bits in place

cupen:
I make a mock case for cluster benchmark.
e.g.: 
```
cd ./examples/cluster-restartgracefully/
make mock-clients-10w
```
But I find a old bug:
The pid cache is not clear when actor.Terminated happened in cluster, I'm looking for why. 

## Reproduce
```
cd ./examples/cluster-restartgracefully/
make mock-clients-10w
# then wait 10~20 seconds after the `actor/grains` terminated.
type `y` and `enter`.
```
image

actor/deadletter.go
```diff
@@ -17,7 +17,11 @@ func NewDeadLetter(actorSystem *ActorSystem) *deadLetterProcess {
 	actorSystem.ProcessRegistry.Add(dp, "deadletter")
 	_ = actorSystem.EventStream.Subscribe(func(msg interface{}) {
 		if deadLetter, ok := msg.(*DeadLetterEvent); ok {
-			plog.Debug("[DeadLetter]", log.Stringer("pid", deadLetter.PID), log.Message(deadLetter.Message), log.Stringer("sender", deadLetter.Sender))
+			plog.Debug("[DeadLetter]", log.Stringer("pid", deadLetter.PID), log.TypeOf("msg", deadLetter.Message), log.Stringer("sender", deadLetter.Sender))
+			// send back a response instead of timeout.
+			if deadLetter.Sender != nil {
+				actorSystem.Root.Send(deadLetter.Sender, &DeadLetterResponse{})
```
rogeralsing, line 23: 👍🏻 good to see that we have this part in Go now.

rogeralsing:
Very very nice job here!

actor/deadletter.go
avinash-getmega: @cupen should we also add some data to `DeadLetterResponse` like PID or message so that the sender has some context to understand why it got this response.

cupen: @bran-san There is only one reason that the sender get a `DeadLetterResponse`, it request to a unreachable actor.

Issue #413, opened 2020-11-03
Update _examples folder to work with the ActorSystem api


rogeralsing:
In this WiP PR https://github.com/AsynkronIT/protoactor-go/pull/412

There are still a lot of examples that haven't been updated to work with the main changes in there.
This could be a good getting started with task for newcomers.

Checkout the branch in the PR and work from there.

rogeralsing:
Some cluster and remote examples still needs updating

umitkavala:
Working on it. I'll update all of them in one go.

umitkavala:
When I try to run example for **cluster-metrics** I saw **pidCache** for cluster is null 
![image](https://user-images.githubusercontent.com/3861934/98468942-062c7d00-21dd-11eb-979e-8193de4cee7a.png)

because we have 2 cluster instances. I've created new one on **protos_protoactor** on my previous commit to use rootContext.
1: in _examples\cluster-metrics\member\main.go
2: _examples\cluster-metrics\shared\protos_protoactor.go
I did following changes. 
![image](https://user-images.githubusercontent.com/3861934/98469002-63c0c980-21dd-11eb-9f7b-b53a25f3db97.png)
and protos_protoactor.go :
```
var c *cluster.Cluster

// SetCluster pass created cluster
func SetCluster(cluster *cluster.Cluster) {
	c = cluster
}
```

What do you think? @rogeralsing @cupen

cupen:
@umitkavala  The grain code was generated by `protobuf plugin`, here is the template code.
https://github.com/AsynkronIT/protoactor-go/blob/f0c77f8866372dc950046033389b7542dd8d00f2/protobuf/protoc-gen-gograin/template.go#L100-L103

I think it could be fixed here.

umitkavala:
Thanks @cupen. I'll do it today.

cupen:
@umitkavala En... Sorry I misunderstood you. It seems that Go is not ready to use grain code generator now. But fortunately, it's not the only way to works with cluster.
Your way is right, go ahead. :smiley:

nightwolfz:
@rogeralsing @umitkavala with the latest changes, the Kinds don't show up as tags in Consul any more

rogeralsing:
@nightwolfz thanks for reporting, I'll take a look asap 👍🏻

PR #454, opened 2021-01-22
Update cluster broadcast example

2m:
This updates cluster broadcast example to use the new `cluster.Call`
api. Also it completes the migration from the consul provider to the
automanaged one.


2m:
This also removes cluster creation in the shared example codebase as suggested by @umitkavala in https://github.com/AsynkronIT/protoactor-go/issues/413#issuecomment-723591334

rogeralsing:
👍🏻 very nice!

PR #456, opened 2021-01-25
Endpoint actor now doesn't crashloop

cupen:
Merged #369, no conflicts now.

PR #457, opened 2021-01-27
Revert "Endpoint actor now doesn't crashloop"

cupen:
Reverts AsynkronIT/protoactor-go#456

Sorry, this PR#456 should be further perfected.

Here is what we talk about it.
image


nightwolfz:
To reproduce the issue, simply use static ports instead of dynamic (0).

Issue #227, opened 2018-03-14
Cross compilation protoactor-go ?


ealoshinsky:
And so on your laptop with the Linux operating system for copying we execute the following command in the terminal:

> CGO_ENABLED = 0 GOOS = windows GOARCH = 386 go build -o test.exe main.go

Excellent assembly is finished, drag test.exe to the machine with Windows ... Trying to run the test.exe we get the runtime error, the output of the error is shown below:

> D:\public>test.exe
> panic: runtime error: invalid memory address or nil pointer dereference
> [signal 0xc0000005 code=0x0 addr=0x0 pc=0x4012dc]
> 
> goroutine 1 [running]:
> sync/atomic.AddUint64(0x11820424, 0x1, 0x0, 0x0, 0x1)
>         /usr/lib/go-1.9/src/sync/atomic/asm_386.s:112 +0xc
> github.com/AsynkronIT/protoactor-go/internal/queue/goring.(*Queue).Push(0x118204
> 20, 0x505200, 0x117d0258)
>         /opt/go/src/github.com/AsynkronIT/protoactor-go/internal/queue/goring/qu
> eue.go:57 +0x24f
> github.com/AsynkronIT/protoactor-go/mailbox.(*unboundedMailboxQueue).Push(0x117d
> 0238, 0x505200, 0x117d0258)
>         /opt/go/src/github.com/AsynkronIT/protoactor-go/mailbox/unbounded.go:13
> +0x33
> github.com/AsynkronIT/protoactor-go/mailbox.(*defaultMailbox).PostUserMessage(0x
> 117f2080, 0x505200, 0x117d0258)
>         /opt/go/src/github.com/AsynkronIT/protoactor-go/mailbox/mailbox.go:56 +0
> x87
> github.com/AsynkronIT/protoactor-go/actor.(*localProcess).SendUserMessage(0x117d
> 8140, 0x11820400, 0x505200, 0x117d0258)
>         /opt/go/src/github.com/AsynkronIT/protoactor-go/actor/local_process.go:1
> 5 +0x36
> github.com/AsynkronIT/protoactor-go/actor.(*PID).Tell(0x11820400, 0x505200, 0x11
> 7d0258)
>         /opt/go/src/github.com/AsynkronIT/protoactor-go/actor/pid.go:45 +0x49
> main.main()
>         /tmp/attempt_actor.go:23 +0x96
> 
> D:\public`

What am I doing wrong?  Who can help in the assembly for different platforms?

[...]

PR #468, opened 2021-04-06
fix: 64-bit alignment of 64-bit words accessed atomically

fermuch:
From `sync/atomic` [docs](https://golang.org/pkg/sync/atomic/#pkg-note-BUG):

> On ARM, 386, and 32-bit MIPS, it is the caller's responsibility to arrange for 64-bit alignment of 64-bit words accessed atomically. The first word in a variable or in an allocated struct, array, or slice can be relied upon to be 64-bit aligned.

Without these changes, `Spawn` gave the following panic on ARM:

```
panic: unaligned 64-bit atomic operation
```

I stumbled upon this while using protoactor on a raspberry pi. These two are the only ones I could find, but I am not well versed in Go nor do I understand enough about protoactor to feel comfortable saying these are all the alignments needed.

### DIFF
--- a/actor/process_registry.go
+++ b/actor/process_registry.go
@@ -7,8 +7,8 @@ import (
 )
 
 type ProcessRegistryValue struct {
-	ActorSystem    *ActorSystem
 	SequenceID     uint64
+	ActorSystem    *ActorSystem
 	Address        string
 	LocalPIDs      cmap.ConcurrentMap
 	RemoteHandlers []AddressResolver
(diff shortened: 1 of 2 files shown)


rogeralsing:
🤯 I did not know this.
cc @cupen  @PotterDai any thoughts on this?

fermuch:
@rogeralsing imagine my face when I saw that! The binary I was using was stripped, so it was even more cryptic! (I got a pointer error from the atomic module itself!!!1!)

It seems like staticcheck has a check for that, but I could not get it to report those fields: https://staticcheck.io/docs/checks#SA1027

cupen:
It looks similar to #227 .

PR #472, opened 2021-04-12
feat: cluster provider by zookeeper

qjpcpu:
add cluster provider of zookeeper


rogeralsing:
Oh, this is super nice! will review in detail

HaileyStorm:
@qjpcpu The zk test seems to hang. I've followed the main readme instructions up to the test point - and actually taken the extra step of using docker-compose on the yml file in etcd so the etcd test doesn't hang - but zk is hanging it. Is there some dependency I need to resolve or process I need to have running in the background?
All tests complete if I run: ``go test `go list ./... | grep -v consul | grep -v zk` | grep -v 'no test files'``
Tests hang after etcd if I run, as shown in the readme: ``go test `go list ./... | grep -v consul` | grep -v 'no test files'``

Thank you!

qjpcpu:
> @qjpcpu The zk test seems to hang. I've followed the main readme instructions up to the test point - and actually taken the extra step of using docker-compose on the yml file in etcd so the etcd test doesn't hang - but zk is hanging it. Is there some dependency I need to resolve or process I need to have running in the background?
> All tests complete if I run: `` go test `go list ./... | grep -v consul | grep -v zk` | grep -v 'no test files' ``
> Tests hang after etcd if I run, as shown in the readme: `` go test `go list ./... | grep -v consul` | grep -v 'no test files' ``
> 
> Thank you!

the zk test wouldn't hang any test because all infra is mocked, looks like your etcd docker doesn't setup correctly, pls check that

qjpcpu:
@HaileyStorm the zk test wouldn't hang any test because all infra is mocked, looks like your etcd docker doesn't setup correctly, pls check that

HaileyStorm:
Ah. I ran the container using `docker-compose up -d`, from the cluster/etcd
directory with the yml file in it, so I didn't even consider that as a
source of the problem honestly. I will investigate. Thank you.

PR #479, opened 2021-05-01
Improve cluster consul provider

chongkai:
The following enhancements are included:
1. Use actor to replace go routines, in order for easier concurrency control, error handling, and scheduling for repeated tasks
2. Use consul's watch API to replace active queries

### DIFF
--- a/cluster/consul/consul_provider.go
+++ b/cluster/consul/consul_provider.go
@@ -14,9 +14,7 @@ import (
 
 var (
 	ProviderShuttingDownError = fmt.Errorf("consul cluster provider is shutting down")
-	// for mocking purposes this function is assigned to a variable
-	blockingUpdateTTLFunc = blockingUpdateTTL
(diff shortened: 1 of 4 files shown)


rogeralsing:
Many thanks!
I will continue to review in more detail 
Great use of Become for the lifecycle of the registrations.

Somewhat skeptical of the `MemberListUpdated` message, from a functionality perspective, it looks like it doesn't add any value over a plain function call from the place where it is sent.

If we add some proper abstractions ontop of this, we could probably use this as the base for all providers going forward.
e.g. switching out the consul API for Kubernetes API in such provider

chongkai:
@rogeralsing , I think using message `MemberListUpdated` could help avoid introducing a synchronous call to the watcher callback, and the actor could become the control center.

And yes, I was also thinking this actor could become a part of the "core" cluster management, and a plugin could be implemented for each provider.

rogeralsing:
Merging this, good job!

PR #487, opened 2021-05-10
Context & Process interface implement add assign statement for compile check

tutumagi:
1. modify readme for latest api.
2. Context & Process interface implement add assign statement for compile check


rogeralsing:
Many thanks for this!

PR #501, opened 2021-08-05
Expose user message queue count in mailbox

prajwalmu:
This PR exposes the User message queue length externally, via the Mailbox interface. We always had this capability in the dotnet sdk of protoactor. This field is useful in case the caller needs to check how many messages are in the user message queue prior to taking appropriate action.

Review by @rogeralsing


rogeralsing:
Very nice, thank you!

Issue #500, opened 2021-07-30
Proto Grain generator generates code for imported files


rogeralsing:
The Protoc code generator we have for grains, incorrectly generates code for imported files also.
This means that if you have a RPC service definition in a.proto, and import b.proto, which does not have any RPC service definition. there will still be a `b` artifact generated. but with no grains in it.

cupen:
You mean `protoc -I=.  --gograinv2_out=.  ./a.proto`? 
or  `protoc -I=.  --gograinv2_out=.  ./*.proto`?

cupen:
Oh, I see it.

PR #504, opened 2021-08-14
fix: issue #500

Tochemey:
https://github.com/AsynkronIT/protoactor-go/issues/500


rogeralsing:
Can you add some context to this one?
This will kill the entire process, so nothing is generated at all, no?
Or is the generator invoked multiple times by protoc here?

Tochemey:
> Can you add some context to this one?
> This will kill the entire process, so nothing is generated at all, no?
> Or is the generator invoked multiple times by protoc here?

@rogeralsing  Valid point. We only generate the grain file for only the proto files that only have service definitions and ignore the rest.

PR #514, opened 2021-10-04
Cluster broadcast

marcinbudny:
This PR adds ability to broadcast events in a similar way the .NET version does. A new function on MemberList allows to pass event to all members of the cluster, that can be subscribed via EventStream.


cluster/member_list.go
```diff
@@ -212,3 +213,13 @@ func groupMembersByKind(members []*Member) map[string][]*Member {
 	}
 	return groups
 }
+
+func (ml *MemberList) BroadcastEvent(message interface{}) {
```
marcinbudny, line 217: This function does not have the `includeSelf` parameter like the .NET counterpart. I don't think the .NET one is implemented correctly, as it compares the member id with actor system id, with former coming from cluster provider, and latter being a generated guid.

rogeralsing: >member id with actor system id

They are the same, the actor system id is the one registered into the cluster provider

cluster/member_list.go
```diff
@@ -212,3 +213,13 @@ func groupMembersByKind(members []*Member) map[string][]*Member {
...
+	ml.mutex.RLock()
+	defer ml.mutex.RUnlock()
+
+	for _, member := range ml.members {
+		pid := actor.NewPID(member.Address(), "eventstream")
+		ml.cluster.ActorSystem.Root.Send(pid, message)
```
rogeralsing, line 223: 👍🏻

cluster/member_list.go
marcinbudny: Correct, I did not follow this deep enough to see that e.g. Kubernetes provider sets pod label to actor system id.

PR #518, opened 2021-11-15
refactor: make use of the consul server address in the configuration

Tochemey:
This PR address an issue when the consul provider. The current behaviour is such that one will have to pass in as an option the consul server address even if it is set in the configuration before the provider can pick it. That will result in `dial tcp 127.0.0.1:8500: connect: connection refused"`. The only way to solve that is to do something like this:

```go
serverAddr := "consul.contoso.com:8600"
consul.NewWithConfig(&api.Config{Address: serverAddr}, consul.WithConsulServerAddress(serverAddr))
```
With this PR one just have to do this:

```go
serverAddr := "consul.contoso.com:8600"
consul.NewWithConfig(&api.Config{Address: serverAddr})
```

### DIFF
--- a/cluster/consul/consul_provider.go
+++ b/cluster/consul/consul_provider.go
@@ -3,9 +3,10 @@ package consul
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/AsynkronIT/protoactor-go/actor"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


rogeralsing:
Very nice!

Issue #519, opened 2021-11-18
Unseteable log levels for all the cluster providers 


DamnWidget:
**Bug Description**

Currently is not possible to set the log level for any of the cluster providers as the `plog` variable is not exported and there is no `SetLogLevel` (or similar) exported function to set it up.

**Affected Packages**
* clustrer/automanaged
* cluster/consul
* cluster/etcd
* cluster/zk

DamnWidget:
Closed by merge #520

PR #520, opened 2021-11-18
fix unseteable loggers for cluster providers #519

DamnWidget:
# Abstract

As reported in #519 the log levels for all cluster providers are impossible to set. This PR adds a `log.go` file to all of the aforementioned providers, moves the `plog` variable creation into it and adds an exported `SetLogLevel` function to the packages


rogeralsing:
Very nice!

Issue #452, opened 2021-01-21
Timer error


huwei123:
**Describe the bug**
timer.go Timer error State cannot get the correct value

`
func startTimer(delay, interval time.Duration, fn func()) CancelFunc {
	var t *time.Timer
	var state int32
	t = time.AfterFunc(delay, func() {
		state := atomic.LoadInt32(&state)   // **Notice that the state variable has become internal.**
		for state == stateInit {
			runtime.Gosched()
			state = atomic.LoadInt32(&state)  //**This value is going to stay the same**
		}

		if state == stateDone {
			return
		}

		fn()
		t.Reset(interval)
	})

	atomic.StoreInt32(&state, stateReady)

	return func() {
		if atomic.SwapInt32(&state, stateDone) != stateDone {
			t.Stop()
		}
	}
}
`

PR #521, opened 2021-11-18
fix: fix infinite loop in startTimer

DamnWidget:
# Abstract

As reported in #452, the function passed to `time.AfterFunc` can get stuck into an infinite loop if the `state` variable from the outer scope is ever equal to `stateInit` or in other words, if the body of function passed to `time.AferFunc` is executed before `atomic.StoreInt32(&state, stateReady)` in the outer scope

```go
func startTimer(delay, interval time.Duration, fn func()) CancelFunc {
	var t *time.Timer
	var state int32
	t = time.AfterFunc(delay, func() {
                 // at this point 'state' is bounded to the local scope 
                 // this will "shadow" the outer scope state var
		state := atomic.LoadInt32(&state) 
                 // we are comparing the local scope version here
		for state == stateInit {  
			runtime.Gosched()
                         // state will never be set with the value 
                         // of the outer scope one as it has been shadow
			state = atomic.LoadInt32(&state)  
		}

		if state == stateDone {
			return
		}

		fn()
		t.Reset(interval)
	})

	// ensures t != nil and is required to avoid data race in
	// AfterFunc calling t.Reset
	atomic.StoreInt32(&state, stateReady)

	return func() {
		if atomic.SwapInt32(&state, stateDone) != stateDone {
			t.Stop()
		}
	}
}
```

Although this is difficult to occur it is not impossible that the Go scheduler, schedules the goroutine running the function in the `AfterFunc` earlier than the rest of the outer body, if that happens, the function will be stuck in that loop forever as it will never be able to update the value of the local `state` variable

## PR Content

* This PR simplifies the loop making impossible for the situation to occur
* It also adds an example on how to use the scheduler package to the `_examples` directory

### DIFF
--- a/_examples/scheduler/go.mod
+++ b/_examples/scheduler/go.mod
@@ -1,0 +1,10 @@
+module scheduler
+
+go 1.16
+
+replace github.com/AsynkronIT/protoactor-go => ../..
+
+require (
+	github.com/AsynkronIT/goconsole v0.0.0-20160504192649-bfa12eebf716
+	github.com/AsynkronIT/protoactor-go v0.0.0-00010101000000-000000000000
+)
--- a/_examples/scheduler/main.go
+++ b/_examples/scheduler/main.go
@@ -1,0 +1,73 @@
+package main
+
+import (
+	"log"
+	"math/rand"
+	"sync"
+	"time"
+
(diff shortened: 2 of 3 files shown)


rogeralsing:
👍🏻

Issue #477, opened 2021-04-29
Metrics plugin


rogeralsing:
Port the metrics plugin from .NET to Go.
We need to find a stable abstraction for metrics that we can use to for this

rogeralsing:
Open Telemetry?
https://github.com/open-telemetry/opentelemetry-go

rogeralsing:
Very much related to https://github.com/AsynkronIT/protoactor-go/issues/466

rogeralsing:
https://github.com/AsynkronIT/protoactor-dotnet/tree/dev/src/Proto.Actor/Metrics

DamnWidget:
I am a bit confused about the wording used in this issue. Shall the OpenTelemetry support be implemented as an implementation of the `plugin.plugin` interface (thus using MiddleWare) or like in the DotNet version (directly filling the instruments in the specific relevant code)?

PR #529, opened 2021-11-28
Metrics plugin addition, solves #477

DamnWidget:
# Abstract
Adds OpenTelemetry support for protoactor-go

## Details
* Mimics .NET implementation
* Metrics live as an Extension for the ActorSystem
* Uses generic OpenTelemetry meters and instrumentation abstractions
* Must be opt-in using `WithDefaultPrometheusProvider` or custom `WithMetricsProviders` 
* Can be used with any provider that implements the OpenTelemetry exporter interface
* Added examples for both Prometheus and Custom provider on `_examples`

Solves #477


rogeralsing:
This is amazingly good! merging

PR #574, opened 2022-02-25
[DO NOT MERGE YET] fix: pass Watch system message to inner.ReceiveDefault

DamnWidget:
# Abstract 

The protoc-gen-gograin boilerplate generation swallows any `SystemMessage`'s that comes to the grains, this is fine in the traditional sense of the grain lifecycle being already managed but in some situations, like with the `Watch` message we might be interested in handle the message in the inner implementation

## (partial?) Solution

I added a small conditional to send the `actor.Watch` and `actor.Unwatch` events to the inner implementation `ReceiveDefault` hook method. We might be interested in also pass other system messages.

## Backward compatibility

The change is backward compatible in the sense that no code built with older versions of `protoc-gen-gograin` is gonna stop working, simply, grain inner implementations that have been compiled with older version boilerplate code will not catch Watch and Unwatch at all that will be completely fine

### DIFF
--- a/protobuf/protoc-gen-gograinv2/template.go
+++ b/protobuf/protoc-gen-gograinv2/template.go
@@ -116,11 +116,12 @@ func (a *{{ $service.Name }}Actor) Receive(ctx actor.Context) {
 		ctx.Poison(ctx.Self())
 
 	case actor.AutoReceiveMessage: // pass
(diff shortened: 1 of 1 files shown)


DamnWidget:
There is something else going on here, I am closing this PR for now while I get some time to investigate the issue

PR #655, opened 2022-04-22
Prevent grain protogen from trying to generate code for includes

orktes:
The grain generator is currently ignoring FileToGenerate passed to by protoc. This means it will try to generate grain code also based on the files that were include in the actual proto file given as input.

So something like

```sh
$ protoc -I ../protos-for-grpc --gograinv2_out=. grain.proto
```
will attempt to generate grains for all the files included in `grain.proto` from the `../protos-for-grpc` include directory.

I ran into multiple issues (https://github.com/asynkron/protoactor-go/issues/648) with the grain protogen with this one being the only one with no viable workaround and also prevents using the workarounds listed for the other issues. So fixing this one first.


rogeralsing:
Very nice!

PR #656, opened 2022-04-22
abc

rogeralsing:


rogeralsing:
colors in Github Actions

cupen:
Maybe squash some commits into one would be nice and clearer.

e.g.: Just `git rebase -i  HEAD~8`  then `push --force` to the PR branch

PR #707, opened 2022-08-12
Add MaxRetryCount field to remote config

rgngl:
This way the `EndpointWriter` doesn't indefinitely try to connect to a non-existing host.


c10h22:
This fix is really needed as we don't want to retry indefinitely to connect to a host that goes offline.
We can make this better by implementing a custom `MaxRetryCount` by `state.address` a kind of callback function that returns an integer based on the parameter `state.address`

nightwolfz:
Very much needed.

Issue #639, opened 2022-04-15
Port heartbeat eviction


rogeralsing:
https://github.com/asynkron/protoactor-dotnet/blob/dev/src/Proto.Cluster/Gossip/Gossiper.cs#L167

rogeralsing:
Closed in #783

Issue #599, opened 2022-04-05
Port over the Cluster PubSub feature from .NET


rogeralsing:
Proto.Actor .NET has a Cluster PubSub feature. this should exist for Go also

rogeralsing:
Done in #783

PR #783, opened 2023-01-21
Port over cluster pubsub, test provider and gossip heartbeat eviction from .net #599 #639

Kunduin:
❤️❤️❤️ This is a very big pr.  

I know it is better not to submit so much code at once, but because cluster pubsub is very complicated so I decided to submit it after all the test cases are passed. Although I have tried to keep the same project structure and naming as .net as much as possible, there are still many interfaces that are not as natural as .net. I will respond to any comment as soon as possible and keep improving this pr. Hopefully this pr will help make the project better.

The cluster pubsub in this pr basically passes all the test cases that appear in the csharp version, the only test case that is missed is `PubSubBatchingProducerTests.Can_handle_publish_timeouts`. This is because the go version of `cluster.Call` does not return nil due to timeouts, while the .net version of `cluster.requestAsync` returns null on timeouts.

In order to test pubsub in the cluster, I also implemented Test Provider, gossip heartbeat eviction, gossip graceful left.

Again, sorry for so much code at once. I would be very happy if the maintainer would help to review this part of the code.


rogeralsing:
Seems to be missing some deps in the pubsub code:

```
=== Errors
Error: cluster/pubsub_producer.go:442:19: undefined: atomic.Bool
Error: cluster/pubsub_producer.go:508:21: undefined: atomic.Bool
Error: cluster/pubsub_producer.go:518:16: undefined: atomic.Bool
Error: cluster/pubsub_producer.go:583:18: undefined: atomic.Bool
```

rogeralsing:
I'll merge and look into the error here :-)

Kunduin:
atomic.Bool seems to be a new feature of go1.19. It may be better to use atomic.int32 or something that go 1.18 support. It is my mistake not to checkout the feat version number 🤣

PR #834, opened 2023-03-26
Ported the ability to cancel long-running tasks in actors

mhelleborg:
### Summary

Ported the `CancellationToken` pattern used in dotnet and exposed `context.Context`  as `Ctx()` on the actor context. 
This should make easier to cancel long running operations in actors gracefully when stopping.


rogeralsing:
There is a pretty drastic perf hit from this:

Before:
Skärmavbild 2023-03-30 kl  17 40 45

After:
Skärmavbild 2023-03-30 kl  17 40 20

PR #855, opened 2023-07-11
fix etcd leaseId does not renew 

tutumagi:
The `etcd leaseId` does not renew in `keepAliveForever`. Because `leaseId` is always greater than zero after `startMember`. 

**Reproduce**: Shut down etcd server, and restart etcd server after 30 seconds. It will print `error: requested lease not found`. 

Fix:
Remove the statement: `if leaseId <= 0 {}`.


tutumagi:
Fix the #853

PR #993, opened 2023-12-19
add reentrancy

qazwsxedckll:
Two remaining problems.

1. Users have to clone and copy `protobuf/protoc-gen-go-grain/options/options.proto`, in order to use options. grpc-gateway also uses this method.

2. Users have to unmarshal `*cluster.GrainResponse` to response type when using reenterancy. I cannot find a place to do this automatically.
```
                f, err := client.InvokeServiceFuture(&InvokeServiceRequest{Name: "Bob"})
		ctx.ReenterAfter(f, func(resp interface{}, err error) {
			if err != nil {
				onError(err)
				return
			}

			switch msg := resp.(type) {
			case *cluster.GrainResponse:
				result := &InvokeServiceResponse{}
				err = proto.Unmarshal(msg.MessageData, result)
				if err != nil {
					onError(err)
				}
				respond(&InvokeServiceResponse{
					Message: req.Name + " from " + ctx.Identity() + " and " + result.Message,
				})
			case *cluster.GrainErrorResponse:
				onError(errors.New(msg.Err))
			default:
				onError(errors.New("unknown response"))
			}
		})
```


rogeralsing:
@qazwsxedckll , we could get rid of GrainResponse completely. see https://github.com/asynkron/protoactor-go/issues/994

it is just a wrapper of the result message and don´t really bring any value

rogeralsing:
This looks fantastic!

qazwsxedckll:
I suddenly realized that the respond and onError callback are not necessary. You can just call `ctx.Respond()` indise `ctx.ReenterAfter()`. I was just imitating the interface in dotnet.

PR #1010, opened 2023-12-30
feat: publish should only accept proto.Message

qazwsxedckll:
If topic actor is on another node, only proto.Message is accepted. Otherwise endpointSupervisor will crash and the node will be blocked.


qazwsxedckll:
@rogeralsing 

I have reproduce the problem [here](https://github.com/qazwsxedckll/protoactor-go/tree/fix-recover-from-non-proto-remote-message/tests).

Just cd to the folder and run `go test -v -tags=integraion .`.

The node seems never become healthy again which I cannot figure out why and I am worried about unhealthy node existing without anyone knowing it.

PR #1014, opened 2024-01-05
fix recover from non proto remote message

qazwsxedckll:


qazwsxedckll:
https://github.com/mgechev/revive/issues/956

qazwsxedckll:
@rogeralsing

PR #1022, opened 2024-01-19
Extend error response

qazwsxedckll:
1. add ErrorReason in proto file.
```
enum ErrorReason {
  USER_NOT_FOUND = 0;
}
```

2. return in service
```
return nil, ErrUserNotFound("not found")
```

3. assert error
```
if IsUserNotFound(err) {
	fmt.Println("user not found")
}
```


cupen:
@qazwsxedckll  Why was the directory name `_examples`  changed back?  See  #384

cupen:
@qazwsxedckll Tracking the `go.sum` file of examples may causing a lot unnecessary changes.

rogeralsing:
👋🏻 @cupen

qazwsxedckll:
@cupen 
It is ignored by go build and I cannot get any lint from vscode even if I add go.work. I understand about removing example dependencies from the main module by adding go.mod to every example. 

Maybe there were many problems before go.work came out so add underscore to ignore them make sense. Now we can just run `go work init && go work use -r .`  to add all examples.

I have no opinions on `go.sum` files.

cupen:
Hi @rogeralsing   @qazwsxedckll  To my point of view, these examples are not a part of `protoactor-go` , but rather some usecases for demonstration purposes. So the `examples` and `protactor-go` can be treated as two separate projects, it might be enough to just place a `go.work` in the `examples` directory for linter.

So, for avoid the lots of unnecessary changes with `go.sum` of `examples`, is there a better way  than `.gitignore`? 

btw: These examples also could be the first choice to check our locally modification, so the `replace xxx => ./..` in `go.mod` of `examples` is useful. And also it could works with `go.work`.

qazwsxedckll:
@cupen [First, `go.work` is not recommanded to be commited to the repository](https://github.com/golang/go/issues/53502). Run `go work init && go work use -r .` in the root directory is the way to go. When using `go.work`, it is alway using the local modifications, no matter the verson chosed in `go.mod`. When not using `go.work`, examples are using the version specified in `go.mod`. I find it is extremely useful. So `replace` is not neccessay here. You can try it out first.

cupen:
@qazwsxedckll Yes, that's true.  I mistakenly thought you wanted to put a `go.work` file in the root path for resolve the linter problems, so I suggested use the `examples` as an alternative. Sorry for that.

PR #1140, opened 2025-02-12
Removes race condition cluster get pid issue 1131

raulriverarojas:
I read the contributing.md and followed the steps
Added tests for manager in disthash
TestManagerConcurrentAccess simulates the race condition by spawning two goroutines - one updates topology continuously while another performs identity lookups, verifying no data races occur under concurrent read/write operations to the rdv field.
TestConcurrentClusterOperations creates a 3-node cluster and performs 100 concurrent identity lookups across different nodes

The tests reveal there is a lot of different data races but no longer in onClusterTopology and get.

### DIFF
--- a/cluster/identitylookup/disthash/manager.go
+++ b/cluster/identitylookup/disthash/manager.go
@@ -1,12 +1,12 @@
 package disthash
 
 import (
-	"log/slog"
-	"time"
-
 	"github.com/asynkron/protoactor-go/actor"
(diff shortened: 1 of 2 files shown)


dshyryayev:
I am curios if this project still alive? or not?

GetSource1234:
yeah, same question here. @rogeralsing do you have any information about the project's current state?

PR #1233, opened 2025-08-25
Handle endpoint connection errors

rogeralsing:
## Summary
- capture error returned from endpoint supervisor connection
- log failures and avoid dereferencing nil endpoint on connect

## Testing
- `make lint`
- `curl -s localhost:8500/v1/status/leader`
- `go test ./cluster/... -count=1`
- `go test ./scheduler -run TestNewTimerScheduler -v`
- `go test ./remote/... -count=1`
- `go test ./persistence/... -count=1`

------
https://chatgpt.com/codex/tasks/task_e_68ac089d7bd483289bb79ac60d4537cd

### DIFF
--- a/remote/endpoint_manager.go
+++ b/remote/endpoint_manager.go
@@ -31,7 +31,11 @@ func (el *endpointLazy) connect() {
 	el.manager.remote.actorSystem.Logger().Debug("connecting to remote address", slog.String("address", el.address))
 	em := el.manager
 	system := em.remote.actorSystem
-	rst, _ := system.Root.RequestFuture(em.endpointSupervisor, el.address, -1).Result()
+	rst, err := system.Root.RequestFuture(em.endpointSupervisor, el.address, -1).Result()
+	if err != nil {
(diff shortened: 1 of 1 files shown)


rogeralsing:
Thanks for tightening up the error path here. One thing I noticed: when  returns an error,  now returns without storing an endpoint.  still does  followed by , so we still panic—just a frame later. We either need to guard that load or store some sentinel before returning, otherwise the failure mode stays the same. Happy to re-test once that's adjusted.

rogeralsing:
Follow-up fix: if the `RequestFuture` to `EndpointSupervisor` fails, we no longer panic on `atomic.Value.Load`/type assert.

- `endpointLazy.Get()` now returns `nil` when connect didn’t store an endpoint.
- `ensureConnected()` drops the failed entry so a later call can retry.
- Call sites handle `nil`: deliver -> DeadLetter, watch -> Terminated(AddressTerminated), terminate -> Terminated(Stopped).

Local verification: `go test ./remote/... -count=1` and `make lint` (revive warnings are pre-existing).

PR #1272, opened 2026-01-09
feat: Add WithEtcdClient method for etcd provider.

8liang:
Add WithEtcdClient method to allow using a custom etcd client for the cluster's etcd provider.

### DIFF
--- a/cluster/clusterproviders/etcd/config.go
+++ b/cluster/clusterproviders/etcd/config.go
@@ -34,10 +34,18 @@ func WithRoleChangedListener(l RoleChangedListener) Option {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


rogeralsing (approved):
LGTM — thanks for adding WithEtcdClient to allow injecting a preconfigured etcd client (TLS/auth/custom dialer, etc.).

Minor (non-blocking): it may be worth noting in the WithEtcdClient doc comment that WithEtcdConfig/cfg is ignored when a client is provided, and clarifying client ownership (provider currently does not Close() the client).

rogeralsing:
Nice, this was the first fully AI evaluated, tested and auto-merged PR :-)

PR #1280, opened 2026-01-17
Feat/request future concurrency

pipiaha:
### Problem
When using `DefaultContext.RequestFuture` with a shared `GrainCallConfig` (obtained via `DefaultGrainCallConfig` function), the `messagePart` in the context faces non-concurrent-safe operations. Each request may have its own parameters, leading to race conditions when multiple requests modify the shared context simultaneously.

### Root Cause
The `DefaultGrainCallConfig` function returns a shared instance of `GrainCallConfig`. When multiple contexts use this shared config for different requests with distinct parameters, they concurrently modify the `messagePart` field, causing data corruption and unpredictable behavior.

### Solution
- Modified the `DefaultGrainCallConfig` to `NewGrainCallOptions` as Request does.


rogeralsing (approved):
Reviewed. Switching RequestFuture to use NewGrainCallOptions avoids shared mutable defaults and should fix concurrent option mutation. CI is green; no issues found.

